PR #50, opened 2020-09-13
Remove space before question mark

skorotkiewicz:
In English, it is always an error. There should be no space between a sentence and its ending punctuation, whether that's a period, a question mark, or an exclamation mark. There should also be no space before a colon, semicolon, or comma.


lambda2:
Thanks !

skorotkiewicz:
@lambda2: It is also a error to put spaces before exclamation mark.

PR #77, opened 2020-12-18
Address some N+1 queries in `SpeciesSerializer`

chancancode:
From looking at the [Skylight dashboard][1], we found some N+1/repeated queries on the `Api::V1::SpeciesController#show` endpoint.

Screen Shot 2020-12-17 at 3 54 55 PM

In the `SpeciesSerializer`, there are a few `has_many` associations that needs to be serialized. Some of these associations then uses serializers that themselves have other `has_many` associations that needs to be serialized. This resulted in the classic [N+1 queries][2] problem.

The usual way to address is to add an `includes` to the query so Rails knows to eager load all the nested associations with a single query. However, in this case, the query is built by `Panko` and there does not appear to be a way to tell `Panko` to do this.

As a workaround, we turned the `has_many` associations to a normal attribute, in which we manually build the query with the preload instruction and serialize them using `ArraySerializer`.

The tests appear to be broken even before this commit, so we are unable to run them against these changes. However, we visited the API endpoint locally and saw no changes to the JSON output, other than a few ordering differences.

[1]: https://oss.skylight.io/app/applications/nz7MAOv6K6ra/recent/6h/endpoints/Api::V1::SpeciesController%23show?responseType=json
[2]: https://www.skylight.io/support/performance-tips#repeated-queries

### DIFF
--- a/app/serializers/species_serializer.rb
+++ b/app/serializers/species_serializer.rb
@@ -155,9 +155,14 @@ class SpeciesSerializer < BaseSerializer
 
   # has_one :plant, serializer: PlantLightSerializer
 
-  has_many :synonyms, serializer: SynonymSerializer
+  # This is causing some N+1 queries, so we turned it into an attribute in order to add the preload
+  # has_many :synonyms, serializer: SynonymSerializer
+  attributes :synonyms
 
-  has_many :foreign_sources_plants, serializer: ForeignSourcesPlantSerializer, name: :sources
+  # This is causing some N+1 queries, so we turned it into an attribute in order to add the preload
+  # has_many :foreign_sources_plants, serializer: ForeignSourcesPlantSerializer, name: :sources
+  attributes :sources
+
   # has_many :species_distributions, serializer: SpeciesDistributionSerializer, name: :distribution
 
   def image_url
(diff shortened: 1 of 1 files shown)


app/serializers/species_serializer.rb
```diff
@@ -189,26 +194,23 @@ def edible
...
+    end
   end
 
   def common_names
     object.common_names.group_by(&:lang).transform_values {|e| e.pluck(:name) }
   end
 
   def distribution
-    object.species_distributions.joins(:zone).group_by(&:establishment).transform_values {|e| e.map {|r| r.zone.name } }
+    object.species_distributions.includes(:zone).group_by(&:establishment).transform_values do |e|
```
chancancode, line 207: This appears to change the ordering/sorting of the zones. We assumed it's not important, because the current code doesn't explicitly specific the sort key, so the ordering is probably indeterministic anyway. If that matters, a `.sort(...)` can be added here to pin down the ordering.

chancancode:
While working on this, we also noticed the [cache](https://github.com/treflehq/trefle-api/blob/d9f40b5740364d19bf5679747699421f95b3fc08/app/controllers/concerns/collection_renderers.rb#L106) for this response has a very low hit rate (less than 1%). We did confirm locally (and on the production API) that the cache _can_ work, and when it does work it does result in significant time-savings.

So, our guess is it's probably a configuration issue where the cache size is way too small for the data set, similar to what [we found on Code Triage](https://blog.skylight.io/using-skylight-to-solve-real-world-performance-problems-part-3-code-triage/) in the past. Unfortunately we don't have access to the information to debug this further, so someone else may with access may have to look into that.

lambda2:
Thanks a lot for this ! 🙏

PR #240, opened 2026-09-03
Explore images: real <img> tags, sized variants, lazy loading, alt text, empty-state placeholder

### LINKED ISSUE #232: Explore images: real <img> tags, sized variants, lazy loading, alt text, empty-state placeholder
The explore grid and species galleries render photos as CSS background-image divs pointing at ORIGINAL-size sources (bs.plantnet.org/image/o/... at 85-360 KB each, 20 per page; some cloudfront JPGs too). Consequences: slow first paint, no lazy loading, zero alt text (half a million species pages invisible to image search), and a species without a photo shows a bare grey rectangle.

Scope:
- Replace background-image divs with real  elements: src using a medium variant (plantnet supports size prefixes, e.g. /image/m/ instead of /image/o/ — verify and map; keep /o/ only for the lightbox/full view), loading="lazy", decoding="async", explicit width/height to prevent CLS.
- alt text: scientific name + common name.
- Empty state: species without an image get a tinted background with a leaf glyph, not raw grey.
- Apply to: explore index cards, species show galleries, genus pages.

Acceptance criteria:
- Explore page image payload drops by an order of magnitude; images lazy-load.
- Every  has meaningful alt; no CLS when images arrive.
- No-image cards look intentional.
- Feature specs green.

Touches: app/views/explore, app/helpers, app/assets/stylesheets (card/gallery rules only)
Blocked by: none (independent of the layout/icon work)

lambda2:
Closes #232

## What

Explore index cards and the species/genus/correction headers rendered
photos as CSS `background-image` divs pointing at full-size plantnet
(`/image/o/...`) or cloudfront sources: no lazy loading, no alt text,
and a bare grey box for species without a photo.

- New `SpeciesPhotoHelper#species_photo_tag` renders a real ``
  with a medium-size plantnet variant (`/image/m/` instead of
  `/image/o/`; non-plantnet URLs are left untouched since their size
  scheme isn't known), `loading="lazy"`, `decoding="async"`, explicit
  `width`/`height`, and `alt` built from scientific + common name.
- Species without a photo now get a tinted placeholder (leaf glyph)
  instead of a grey rectangle.
- Applied to: explore species grid cards, species show header, genus
  show header, and the record-correction header (they all shared the
  same background-image markup/CSS, so all three had to move together).
- `Explore::GenusController#show` now picks the genus' best-ranked
  species regardless of whether it has a photo (previously it only
  considered species *with* a photo, so a genus where none of its
  species had one would 500 on the header partial). This also makes the
  new empty-state placeholder reachable on genus pages.

## Scope note

The species show page has an empty, unimplemented "Images" gallery
section (`id="images"`) that renders no photos today — building a real
multi-photo gallery there is new feature work beyond swapping existing
background-image divs for `` tags, so it's left untouched. The
species header photo is the only existing "gallery" rendering today and
is covered.

## Testing

- New `spec/helpers/species_photo_helper_spec.rb`: URL size-variant
  mapping, alt text, and the img/placeholder branches.
- Extended `spec/requests/web/explore_spec.rb`: asserts real ``
  markup (lazy, sized, alt) and the leaf placeholder on the explore
  index, species show, and genus show pages, and confirms no
  `background-image` remains in any of their responses.
- `bundle exec rspec`: 729 examples, 0 failures (10 pre-existing
  pending specs, unrelated).
- `bundle exec brakeman`: 0 warnings.

Touches: app/views/explore, app/helpers, app/assets/javascripts/sections, app/controllers/explore/genus_controller.rb

### DIFF
--- a/app/assets/javascripts/sections/_explore.scss
+++ b/app/assets/javascripts/sections/_explore.scss
@@ -169,12 +169,9 @@
     min-height: 150px;
     height: 100%;
     display: flex;
-    background-color: #ccc;
     margin: 10px 0;
     border-radius: 3px;
-    background-position: center;
-    background-repeat: no-repeat;
-    background-size: cover;
+    overflow: hidden;
   }
 
   main {
@@ -186,6 +183,32 @@
     p {
       color: #6f6f6f
     }
+  }
+}
+
+// Real <img>/placeholder rendered by SpeciesPhotoHelper#species_photo_tag, replacing
+// the background-image divs the cards and headers used to rely on (see issue #232).
+.species-grid-item-photo,
+.species-correction-image-photo {
+  width: 100%;
+  height: 100%;
+  object-fit: cover;
+  object-position: center;
+  display: block;
+}
+
+.species-grid-item-photo--empty,
+.species-correction-image-photo--empty {
+  width: 100%;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  background-color: #e3f3e6;
+
+  i {
+    color: #3ca35a;
+    font-size: 2rem;
   }
 }
 
@@ -259,22 +282,22 @@
   .species-correction-image {
     height: 100%;
     display: flex;
-    background-color: #ccc;
     margin: 10px 0;
     border-radius: 3px;
-    background-position: center;
-    background-repeat: no-repeat;
+    overflow: hidden;
   }
 
   .species-side-image {
     min-height: 300px;
+    background-color: #ccc;
+    background-position: center;
+    background-repeat: no-repeat;
     background-size: cover;
   }
 
   .species-correction-image {
     min-height: 100px;
     margin: 0;
-    background-size: cover;
   }
 
   .done-footer {
--- a/app/controllers/explore/genus_controller.rb
+++ b/app/controllers/explore/genus_controller.rb
@@ -17,7 +17,10 @@ class Explore::GenusController < Explore::ExploreController
     @page_description = "#{@genus.name} is a genus of the #{@genus.family&.name} family"
     @page_keywords    = [@genus.name, 'genus', 'plant', 'explore'].compact.join(', ')
 
-    @species = @genus.species.where.not(main_image_url: nil).order(wiki_score: :desc).first
+    # Picking the best-ranked species regardless of whether it has a photo lets the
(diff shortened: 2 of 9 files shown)


lambda2:
## Review: Pass

**Contract** (#232): real `` tags with medium-size plantnet variant, `loading=lazy`, `decoding=async`, explicit width/height, meaningful alt text, tinted leaf placeholder for no-photo species, applied to explore index cards + species/genus/correction headers. Scope note excluding the still-unimplemented species "Images" gallery section is reasonable — there's nothing there today to convert.

**CI**: all three gates green — Brakeman, RSpec, RuboCop (`gh pr checks 240`).

**Local verification**: checked out the branch in an isolated worktree and re-ran the affected suite directly (`spec/helpers/species_photo_helper_spec.rb`, `spec/requests/web/explore_spec.rb`): 22 examples, 0 failures. Targeted RuboCop on the changed Ruby files (`genus_controller.rb`, `species_photo_helper.rb`, both specs): 0 offenses.

**Code read**:
- `SpeciesPhotoHelper#species_photo_tag` correctly branches on `main_image_url` presence, swaps `/image/o/` → `/image/m/` for plantnet URLs only (non-plantnet URLs pass through untouched), builds alt from scientific + common name, sets `loading`/`decoding`/`width`/`height`.
- `Explore::GenusController#show` fix (dropping `.where.not(main_image_url: nil)`) correctly resolves the pre-existing 500 when a genus's species all lack a photo, and is covered by the new "leaf placeholder in the genus header" request spec — verified this is a real fix, not just a happy-path test.
- Grepped for stray `background-image` usage in the touched view tree — only `management/species_images/chaos.html.erb` remains, which is out of scope (admin back-office, not `explore/`).
- SCSS split (`.species-side-image` vs `.species-correction-image`) is consistent with the new markup; `.species-correction-image` now relies on `overflow: hidden` to clip the ``/placeholder child instead of `background-size: cover`.

No blocking findings. Nits only, not gating: the empty-state `<i class="fa-leaf">` placeholder has no accessible name (not required by the acceptance criteria, but a `<span class="sr-only">No photo available</span>` would be a nice follow-up), and the plantnet size-variant assumption (`o`/`m`/`s`) is trusted rather than verified against a live response — reasonable given it's documented as PlantNet's public API convention and is covered by unit tests either way.

PR #241, opened 2026-09-03
Expose RateLimit-* headers on every /api response

### LINKED ISSUE #209: Expose rate-limit headers on API responses
rack-attack throttles /api requests (60/min, 600/min for some plans) but clients only learn about it when they hit a 429 (`Retry-After` is already enabled). Well-behaved clients cannot self-regulate.

Scope: enable rack-attack's throttled response headers so every throttled-tracked response carries `X-RateLimit-Limit`, `X-RateLimit-Remaining` and `X-RateLimit-Reset` (or the `RateLimit-*` names, pick one convention and document it). Document the headers in the docs guide about rate limiting.

Acceptance criteria:
- A request spec asserts the three headers on an /api/v1 response and on a 429.
- Header values are consistent with the configured limits (60 baseline).
- No behavior change to the limits themselves.

Touches: config/initializers/rack_attack.rb, spec/requests (new file)
Blocked by: none

lambda2:
Closes #209.

## What

rack-attack throttles `/api` (60 req/min baseline, 600/min for sponsor tokens) but only added `RateLimit-Limit` / `RateLimit-Remaining` / `RateLimit-Reset` headers to the 429 responses it generated when blocking a request. Well-behaved clients had no way to see how close they were to the limit before getting cut off.

- Extracted the header-building logic (already used by `throttled_responder`) into `Rack::Attack.rate_limit_headers`.
- Added `RateLimitHeadersMiddleware`, a small Rack middleware wrapping the whole stack (`unshift`, so it always sees Rack::Attack's 429s too) that reads rack-attack's `env['rack.attack.throttle_data']` -- set on every /api request it evaluates against the throttle, whether or not that request is actually throttled -- and attaches the same three headers to every response.
- No change to the throttle configuration itself (still 60/min baseline, 600/min for `spo-` sponsor tokens).

Defined the middleware inline in the initializer rather than under `lib/` -- referencing an autoloadable `lib/` constant directly inside `config/initializers` raises `NameError`, since Zeitwerk's main autoloader isn't set up yet at that point in boot.

Also updated the *Rate limiting* section of the getting-started guide in `treflehq/documentation` (not pushed yet -- that repo currently has a large amount of unrelated pre-existing local changes, so I left mine uncommitted there rather than mixing them into a commit; flagging separately) to document the three headers and correct the stated limit (120/min in the doc vs. the actual 60/min in code).

## Testing

- New request spec `spec/requests/api/v1/rate_limit_headers_spec.rb`: asserts the three headers on a normal 200 response, that `RateLimit-Remaining` decrements across requests, and that all three headers are present and consistent on the 429 once the limit is exceeded.
- `bundle exec rspec spec/requests/api`: 83 examples, 0 failures.
- `bundle exec rubocop`: 370 files inspected, no offenses.
- Pre-existing, unrelated: `spec/requests/web/{home,explore,management}_spec.rb` fail with 500s on this baseline (`master` @ 61ca485) independent of this change -- confirmed by stashing and re-running.

Touches: config/initializers/rack_attack.rb, spec/requests/api/v1

### DIFF
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -1,10 +1,29 @@
 class Rack::Attack
 
+  REQUEST_PER_IP_THROTTLE = 'request per ip'.freeze
+
   def self.token(req)
     [
       req.env['HTTP_AUTHORIZATION'].to_s.downcase.gsub(/bearer /, ''),
       req.params['token'].to_s.downcase.gsub(/\s+/, '')
     ].reject(&:blank?).first
+  end
+
+  # Builds the RateLimit-* headers (draft IETF RateLimit-Limit /
+  # RateLimit-Remaining / RateLimit-Reset convention -- documented alongside
+  # the getting-started guide's rate limiting section). Shared by the
+  # throttled_responder below (429s) and by RateLimitHeadersMiddleware
+  # (every other /api response).
+  def self.rate_limit_headers(match_data)
+    now = match_data[:epoch_time]
+    reset_at = now + (match_data[:period] - now % match_data[:period])
+    remaining = [match_data[:limit] - match_data[:count], 0].max
+
+    {
+      'RateLimit-Limit' => match_data[:limit].to_s,
+      'RateLimit-Remaining' => remaining.to_s,
+      'RateLimit-Reset' => reset_at.to_s
+    }
   end
 end
 
@@ -16,7 +35,7 @@ end
 limit_proc = proc { |req| Rack::Attack.token(req)&.starts_with?('spo-') ? 600 : 60 }
 
 
-Rack::Attack.throttle('request per ip', limit: limit_proc, period: 60) do |req|
+Rack::Attack.throttle(Rack::Attack::REQUEST_PER_IP_THROTTLE, limit: limit_proc, period: 60) do |req|
   [req.ip, Rack::Attack.token(req)].join('-') if req.path.starts_with?('/api')
 end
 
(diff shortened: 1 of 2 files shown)


lambda2:
🔎 claim-1788457024-5754

lambda2:
## Review: Pass

**Contract** (#209): every /api response — not just 429s — now carries `RateLimit-Limit`/`RateLimit-Remaining`/`RateLimit-Reset`, values consistent with the configured 60/600 limits, throttle config itself unchanged. Docs update is being done separately in the `documentation` repo (cross-repo, transparently flagged in the PR body) — reasonable, nothing in this repo's contract requires it to land in the same PR.

**CI**: all three gates green — Brakeman, RSpec, RuboCop (`gh pr checks 241`).

**Local verification**: checked out the branch in an isolated worktree.
- `bundle exec rspec spec/requests/api/v1/rate_limit_headers_spec.rb`: 3 examples, 0 failures.
- `bundle exec rspec spec/requests/api`: 83 examples, 0 failures — matches the PR's reported count.
- `bundle exec rubocop config/initializers/rack_attack.rb spec/requests/api/v1/rate_limit_headers_spec.rb`: 0 offenses.
- Initially saw 29/49 failures on `spec/requests/web/{home,explore,management}_spec.rb` in the worktree, which looked like a real regression — traced it to a `Sprockets::ArgumentError (link_tree argument must be a directory)`, caused by my worktree missing built JS assets (`git worktree` doesn't carry over gitignored `app/assets/builds`). After `yarn install && yarn build` in the worktree, all 49 pass — confirms this was a local setup artifact, not something introduced by the PR.

**Code read**:
- Verified against the installed `rack-attack` 6.6.1 source: `Throttle#matched_by?` annotates `env['rack.attack.throttle_data'][name]` with `{discriminator, count, period, limit, epoch_time}` for *every* request the throttle block evaluates (truthy discriminator), whether or not it ends up throttled — not only on the matched/blocking path. So `RateLimitHeadersMiddleware` reading that key for non-throttled requests is well-founded, and `Rack::Attack.rate_limit_headers`'s field names (`:count`, `:limit`, `:period`, `:epoch_time`) match that shape exactly.
- `unshift` placement is correct regardless of init order: rack-attack's own railtie only ever `use`s (appends), so whichever initializer runs first, `RateLimitHeadersMiddleware` ends up outermost and always sees Rack::Attack's response (including 429s).
- Reset/remaining math is an exact extraction of the pre-existing 429-path logic into `Rack::Attack.rate_limit_headers`, just generalized with `[limit - count, 0].max` instead of a hardcoded `'0'` — same result on the throttled path, more accurate (decrementing) on the normal path.
- Non-`/api` paths never populate `throttle_data` (the throttle's own discriminator block returns `nil` for those), so the middleware correctly adds no headers there — no risk to `explore/`/`management/` responses, confirmed by the web specs above.
- Diff stays exactly within the PR's stated `Touches:` scope — no unrelated changes.

No blocking findings.

PR #227, opened 2026-09-03
CI hardening: coverage floor, scheduled dependency audit, README badges

lambda2:
Closes #208

- SimpleCov: enforce `minimum_coverage 65` so a PR dropping line coverage below the current baseline (65.9%) fails the RSpec job. The floor is raised as #204 progresses, not here.
- New `.github/workflows/dependency-audit.yml`: weekly scheduled + on-PR (paths: `Gemfile.lock`, `yarn.lock`) job running `bundler-audit` (blocking) and `yarn audit --level high` (non-blocking, `continue-on-error`) so vulnerable dependencies surface visibly in Actions instead of silently between Dependabot PRs.
- README: added CI status, license, and latest-release badges, no other content changes.

Verification:
- Full RSpec suite: 700 examples, 0 failures, 10 pending (pre-existing), coverage 65.94% — matches the issue's stated baseline.
- RuboCop: 368 files inspected, no offenses.
- Brakeman: 0 warnings, 0 errors.
- Smoke-tested `bundler-audit check --update` and `yarn audit --level high` directly against this repo's lockfiles to confirm the commands are correct for this project's toolchain (classic Yarn 1 lockfile format) before wiring them into the workflow.

Note: bundler-audit currently reports real vulnerabilities in this repo's Gemfile.lock (rack, rails-html-sanitizer, websocket-driver) and yarn audit reports several in yarn.lock. That's expected — this PR's job is to make that visible, not to fix them; the scheduled job will run red until those are bumped in follow-up PRs.

Touches: spec/spec_helper.rb, .github/workflows/, README.md

PR #246, opened 2026-09-04
Enable CORS on the public API

### LINKED ISSUE #91: CORS error when calling the API from the client side with a JWT
Hi, and thanks for your work with this API!

I'm trying to build a small application and, bear with me, since I'm a total back-end newbie I'm having a hard time figuring out how to get through CORS issues.

Basically, I can correctly obtain my JWT via the claim API; but, when I use that JWT in my front-end, I always hit the CORS roadblock:

```
Access to XMLHttpRequest at 'https://trefle.io/api/v1/plants/search?token=<token>&q=mint' (redirected from 'http://trefle.io/api/v1/plants/search?token=<token>&q=mint') from origin 'http://localhost:3000' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.
```

I'm also referencing an existing comment because it describes my case:

> I have the same issue but when calling the API from the client side with a fresh JWT token using Fetch:
> 
> ```
> fetch(`https://trefle.io/api/plants?q=rosemary?token=${JWT_TOKEN}`);
> ```
> 
> Results in:
> 
> ```
> Access to fetch at 'https://trefle.io/api/plants?q=rosemary?token={JWT_TOKEN}' from origin
>  'http://localhost:8080' has been blocked by CORS policy: No 'Access-Control-Allow-Origin'
>  header is present on the requested resource. If an opaque response serves your needs,
>  set the request's mode to 'no-cors' to fetch the resource with CORS disabled.
> ```
> 
> I tried fetching passing options, even with mode: 'no-cors', but returns a 401: 
> 
> ```
> fetch(`https://trefle.io/api/plants?q=${query}?token=${JWT_TOKEN}`,
>       {
>         method: 'GET',
>         'Content-Type': 'application/json',
>         'Access-Control-Allow-Origin': '*',
>       }
>  );
> ```

_Originally posted by @lewislbr in https://github.com/treflehq/trefle-api/issues/23#issuecomment-525191479_

I'm totally out of ideas and unable to use your awesome API. Hope you can help.

Thanks!

PS

Sorry if I'm not using the issue template, I don't think this request fits in the categories.

### LINKED ISSUE #211: Enable CORS on the public API
Browser apps calling the API directly get CORS failures (community report: #91). Yet client-side use is clearly intended: JWTs are claimed per-origin (`POST /api/auth/claim` takes an `origin` param and tokens are origin-bound). The `rack-cors` gem and its initializer are scaffolded but commented out.

Scope:
- Add `rack-cors`, configure: allow all origins for `GET`/`HEAD`/`OPTIONS` on `/api/*`, plus `POST /api/auth/claim`. Expose the rate-limit headers (see #209) in `Access-Control-Expose-Headers`.
- Request specs for the preflight (OPTIONS returns the CORS headers) and for a cross-origin GET.
- Documentation note in the client-side-apps guide (docs repo has a page about this — a follow-up docs PR is fine, link it).

The PR should close #91.

Acceptance criteria:
- OPTIONS preflight on /api/v1/plants returns 200 with `Access-Control-Allow-Origin`.
- GET responses on /api carry the header; non-API routes (management, explore, Devise) are untouched.
- Full suite green.

Touches: Gemfile, Gemfile.lock, config/initializers/cors.rb, spec/requests
Blocked by: #210 (Gemfile chain)

lambda2:
Closes #211. Closes #91.

Browser apps calling the API directly get CORS failures (community report #91) even though client-side use is clearly intended — JWTs are already origin-bound via `POST /api/auth/claim`. `rack-cors` was scaffolded (`config/initializers/cors.rb`) but commented out.

## What changed

- Uncommented `rack-cors` in the Gemfile.
- Configured `config/initializers/cors.rb`: all origins allowed for `GET`/`HEAD`/`OPTIONS` on `/api/*`, plus `POST`/`OPTIONS` on `/api/auth/claim`. The `RateLimit-*` headers (#241) are exposed via `Access-Control-Expose-Headers` so well-behaved browser clients can read them.
- Left the existing hand-rolled CORS logic in `Api::ApiController` (`cors_preflight_check` / `cors_set_access_control_headers`) untouched — it still covers the write endpoints (corrections, reports, etc.) that aren't in this issue's scope, and the origin-binding security check on JWTs (`check_jwt!`) is unrelated to CORS headers (that's an app-level check, not something a browser enforces).
- Added `spec/requests/api/v1/cors_spec.rb`: preflight on `/api/v1/plants` and `/api/auth/claim`, a cross-origin GET carrying `Access-Control-Allow-Origin`, exposed `RateLimit-*` headers, and a check that non-API routes are untouched.

## Follow-up

Docs update in the docs repo (client-side-apps guide) — left for a follow-up docs PR as scoped in the issue.

## Testing

- `bundle exec rspec spec/requests/api`: 88 examples, 0 failures (baseline was 83/0 before this change).
- `bundle exec rspec spec/requests`: 143 examples, 34 pre-existing failures (same as the pre-change baseline — all in `spec/requests/web/*`, unrelated to this change; environmental, not something this PR should fix).
- `bundle exec rubocop`: 375 files inspected, no offenses.
- `bundle exec brakeman -q --no-pager`: 0 security warnings.

Touches: Gemfile, Gemfile.lock, config/initializers/cors.rb, spec/requests

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,7 +24,7 @@ gem 'friendly_id', '>= 5.2.4'
 # Use ActiveStorage variant
 # gem 'mini_magick', '>= 4.8'
 gem 'oj'
-# gem 'rack-cors', require: 'rack/cors'
+gem 'rack-cors', require: 'rack/cors'
 
 # API
 # gem 'api-pagination'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -391,6 +391,8 @@ GEM
     rack (2.2.24)
     rack-attack (6.6.1)
       rack (>= 1.0, < 3)
+    rack-cors (2.0.2)
+      rack (>= 2.0.0)
     rack-protection (3.2.0)
       base64 (>= 0.1.0)
       rack (~> 2.2, >= 2.2.4)
@@ -684,6 +686,7 @@ DEPENDENCIES
   pry
   puma
   rack-attack (~> 6.6.1)
+  rack-cors
   rails (~> 8.0.0)
   rails-controller-testing
   recaptcha
--- a/config/initializers/cors.rb
+++ b/config/initializers/cors.rb
@@ -13,6 +13,14 @@
 # boundary). Everything else under /api (writes: corrections, reports...)
 # is intentionally left alone here and keeps relying on the existing
 # Api::ApiController#cors_preflight_check / #cors_set_access_control_headers.
+#
+# rack-cors's Resource#match? (used for BOTH preflight and actual requests)
+# only consults matches_path? and if_proc -- the `methods:` list below is
+# only enforced during preflight (Resource#process_preflight). Without the
+# `if:` guard, an actual cross-origin write request (POST /report, etc.)
+# would still get matched by this catch-all and stamped with CORS headers,
+# regardless of its verb. The `if:` proc is what actually keeps writes out.
+read_methods = %w[GET HEAD OPTIONS].freeze
 rate_limit_headers = %w[RateLimit-Limit RateLimit-Remaining RateLimit-Reset].freeze
 
 Rails.application.config.middleware.insert_before 0, Rack::Cors do
@@ -30,6 +38,7 @@ Rails.application.config.middleware.insert_before 0, Rack::Cors do
     resource '/api/*',
              headers: :any,
              methods: %i[get head options],
+             if: ->(env) { read_methods.include?(env['REQUEST_METHOD']) },
              expose: rate_limit_headers
   end
 end
--- a/config/initializers/cors.rb
+++ b/config/initializers/cors.rb
@@ -5,12 +5,31 @@
 
 # Read more: https://github.com/cyu/rack-cors
 
-# Rails.application.config.middleware.insert_before 0, Rack::Cors do
-#   allow do
-#     origins "example.com"
-#
-#     resource "*",
-#       headers: :any,
-#       methods: [:get, :post, :put, :patch, :delete, :options, :head]
-#   end
-# end
+# Client-side apps are expected to call the read endpoints directly from the
+# browser (community report: #91) and to claim an origin-bound JWT via
+# POST /api/auth/claim (see Api::Auth::AuthController#claim, and
+# Api::ApiController#check_jwt! for the origin check enforced on that token
+# afterwards -- CORS headers are a browser-side courtesy, not the security
+# boundary). Everything else under /api (writes: corrections, reports...)
+# is intentionally left alone here and keeps relying on the existing
+# Api::ApiController#cors_preflight_check / #cors_set_access_control_headers.
+rate_limit_headers = %w[RateLimit-Limit RateLimit-Remaining RateLimit-Reset].freeze
+
+Rails.application.config.middleware.insert_before 0, Rack::Cors do
+  allow do
+    origins '*'
+
+    # More specific than the /api/* resource below -- rack-cors matches the
+    # first resource whose path fits, so this one has to come first or the
+    # catch-all would shadow it and preflight POST requests would be denied.
+    resource '/api/auth/claim',
+             headers: :any,
+             methods: %i[post options],
+             expose: rate_limit_headers
+
+    resource '/api/*',
+             headers: :any,
+             methods: %i[get head options],
+             expose: rate_limit_headers
+  end
+end
--- a/spec/requests/api/v1/cors_spec.rb
+++ b/spec/requests/api/v1/cors_spec.rb
@@ -48,6 +48,16 @@ RSpec.describe 'CORS on the public API', type: :request do
 
       expect(response.headers['Access-Control-Allow-Origin']).to be_nil
     end
+
+    it 'does not add CORS headers to a cross-origin POST to a write endpoint' do
+      plant = create(:plant)
+
+      post "/api/v1/plants/#{plant.id}/report",
+           params: { reason: 'test' },
+           headers: { 'Origin' => origin, 'Authorization' => "Bearer #{user.token}" }
+
+      expect(response.headers['Access-Control-Allow-Origin']).to be_nil
+    end
   end
 
 end
--- a/spec/requests/api/v1/cors_spec.rb
+++ b/spec/requests/api/v1/cors_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe 'CORS on the public API', type: :request do
+
+  let(:user) { create(:user) }
+  let(:origin) { 'https://example.com' }
+
+  describe 'preflight' do
+    it 'answers an OPTIONS preflight for a GET /api/v1 endpoint with the CORS headers' do
+      options '/api/v1/plants', headers: {
+        'Origin' => origin,
+        'Access-Control-Request-Method' => 'GET'
+      }
+
+      expect(response).to have_http_status(:ok)
+      expect(response.headers['Access-Control-Allow-Origin']).to eq('*')
+      expect(response.headers['Access-Control-Allow-Methods']).to include('GET')
+    end
+
+    it 'answers an OPTIONS preflight for POST /api/auth/claim with the CORS headers' do
+      options '/api/auth/claim', headers: {
+        'Origin' => origin,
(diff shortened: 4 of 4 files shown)


lambda2:
🔎 Review

**Checked:** CI (`gh pr checks 246`), diff against #211/#91's acceptance criteria, and the `rack-cors` 2.0.2 source to confirm actual matching behavior (not just reading the initializer). Ran `bundle exec rspec spec/requests/api/v1/cors_spec.rb` in a clean worktree off this branch: 5/5 pass. Also wrote two throwaway probe specs (not committed) to exercise behavior the existing spec doesn't cover.

**Verdict: needs-fixes**

### 1. Write endpoint responses are not actually excluded from CORS (config/initializers/cors.rb)

The PR description says the `/api/*` resource is scoped to `GET/HEAD/OPTIONS` and that write endpoints (corrections, `/report`, ...) are "intentionally left alone" and still rely on `Api::ApiController#cors_set_access_control_headers`. That's not what happens.

`rack-cors`'s `Resource#match?` — used for actual (non-preflight) requests via `Rack::Cors#process_cors` — only checks `matches_path?`; it never consults `methods`. The `methods:` list is only enforced inside `process_preflight` (i.e. it gates the *preflight* response for non-simple requests). For an actual request, whatever resource's path matches gets its headers stamped unconditionally, regardless of HTTP verb.

Repro (cross-origin POST to a write endpoint, with a plain Bearer token — no JWT, so `cors_set_access_control_headers`'s `if @jwt` guard doesn't fire and nothing overrides the middleware's header):

```ruby
post "/api/v1/plants/#{plant.id}/report",
     params: { reason: 'test' },
     headers: { 'Origin' => 'https://evil.example', 'Authorization' => "Bearer #{user.token}" }
# => Access-Control-Allow-Origin: "*"
# => Access-Control-Allow-Methods: "GET, HEAD, OPTIONS"   (declared methods, irrelevant to the actual response)
```

So any regular-token client's cross-origin write response (corrections, `/report`, etc.) is readable by any origin, same as GET — this widens the CORS surface beyond what the PR (and its own code comment) states, for any client that isn't using a JWT-bound token. Given the API's non-cookie bearer-token auth this isn't a CSRF hole, but it is a real gap between the stated design and actual behavior, and it does relax response confidentiality on write endpoints cross-origin.

Fix: gate the `/api/*` resource so it doesn't apply to actual write requests either — e.g. an `if:` proc checking `env['REQUEST_METHOD']` against the allowed set, or splitting reads into their own explicit resource(s) and leaving writes unmatched by rack-cors entirely (so they keep going through only the hand-rolled `ApiController` logic, as intended). Then either add a request spec that pins "no CORS header on a cross-origin POST to a write endpoint," or fix the code comment/PR description if leaving it as-is is actually the intended tradeoff — right now the comment asserts a guarantee the code doesn't provide.

### 2. CI: `bundler-audit` is red (Dependency audit / `bundler-audit`, https://github.com/treflehq/trefle-api/actions/runs/33852743453/job/100959002128)

Fails on pre-existing High-severity CVEs already in `Gemfile.lock` — `puma` 7.2.0 (CVE-2026-47736, CVE-2026-47737), `websocket-driver` 0.8.1 (CVE-2026-61666), plus several `oj` CVEs. Confirmed these aren't introduced by this PR: `git diff master...pr-246 -- Gemfile.lock` only adds `rack-cors (2.0.2)`, nothing touches `puma`/`oj`/`websocket-driver`. `dependency-audit.yml` was added in #227 and only triggers on PRs touching `Gemfile.lock`/`yarn.lock` (plus a weekly cron) — this is the first such PR since, which is why it's surfacing now rather than on master.

Not this PR's fault and out of scope for the CORS work — recommend a separate PR bumping `puma`/`websocket-driver`/`oj` to patched versions, landed independently. Flagging so it doesn't get lost, but the fix belongs elsewhere. (`yarn audit` is `continue-on-error: true`, so its failure isn't blocking — no action needed there.)

Everything else checks out: `RSpec`/`RuboCop`/`Brakeman` green, resource ordering in the initializer is correct and necessary (verified `rack-cors`'s `Resources#match_resource` returns the *first* matching resource, so `/api/auth/claim` has to precede the `/api/*` catch-all or its preflight POST would be denied), exposed `RateLimit-*` header names match #241's actual header names, and non-API routes are untouched per the new spec.

lambda2:
🔧 claim-1788510770-25639

lambda2:
Fixed per the review:

**#1 — write endpoints not actually excluded from CORS.** Confirmed the root cause: `Rack::Cors::Resource#match?` (used for both preflight *and* actual requests) only checks `matches_path?` and `if_proc` — the `methods:` list is only enforced inside `process_preflight`. So the `/api/*` resource was matching and stamping CORS headers on actual write requests too, regardless of verb.

Fix (663f9c8): added an `if:` proc to the `/api/*` resource restricting it to `GET/HEAD/OPTIONS`, so actual requests are gated the same way preflight already was. Added a request spec pinning "no CORS headers on a cross-origin POST to a write endpoint" (`POST /api/v1/plants/:id/report`).

**#2 — bundler-audit CVEs.** Confirmed pre-existing (puma/websocket-driver/oj, unrelated to this PR's `Gemfile.lock` diff) and out of scope here as noted in the review — left for a separate dependency-bump PR.

`bundle exec rspec spec/requests/api/v1/cors_spec.rb` (7 examples) and `bundle exec rubocop` on the touched files are green.

lambda2:
🔎 claim-claim-1788511010-15766

lambda2:
🔎 Review (round 2)

**Checked:** CI (`gh pr checks 246`), the fix commit (663f9c8) against round-1 finding #1, `rack-cors` 2.0.2 source (`Resource#match?`/`Resources#match_resource`/`Cors#process_cors`) to confirm the `if:` proc is actually consulted on actual (non-preflight) requests, and re-ran the full RSpec CI log.

**Verdict: Pass**

### Fix verified

The `if: ->(env) { read_methods.include?(env['REQUEST_METHOD']) }` guard on the `/api/*` resource closes the gap: `Resources#match_resource` calls `r.match?(path, env)` for both `process_preflight` and `process_cors` (the actual-request path), and `Resource#match?` is `matches_path? && (if_proc.nil? || if_proc.call(env))`. For an actual POST, `env['REQUEST_METHOD']` is `'POST'`, so the resource no longer matches and rack-cors stamps nothing — the response falls through to `Api::ApiController#cors_set_access_control_headers`, unchanged, which only sets headers when `@jwt` is present. Non-JWT bearer-token writes now correctly get no cross-origin headers at all, matching pre-PR behavior for that case.

Ran the exact round-1 repro in a clean worktree off 663f9c8: `POST /api/v1/plants/:id/report` cross-origin with a plain bearer token now returns no `Access-Control-Allow-Origin`. The new spec (`does not add CORS headers to a cross-origin POST to a write endpoint`) pins this. `bundle exec rspec spec/requests/api/v1/cors_spec.rb`: 7 examples, 0 failures locally.

Checked one narrow residual: `/api/auth/claim` has no `if:` guard, so a stray (routeless) `GET /api/auth/claim` would still get CORS-stamped on its 404 — but that path only ever routes `POST`, there's no GET handler with a response body to leak, so this isn't a version of the same bug. Not blocking.

### CI

`gh pr checks 246`: RuboCop pass, Brakeman pass, RSpec pass (743 examples, 0 failures, 10 pending — full suite, not just the new spec). `bundler-audit` and `yarn audit` are still red, but as established in round 1: pre-existing High CVEs (puma/websocket-driver/oj) untouched by this PR's `Gemfile.lock` diff (only adds `rack-cors 2.0.2`), `yarn audit` is `continue-on-error: true`, and neither is one of the three required jobs (RSpec/RuboCop/Brakeman) this repo's CI contract gates on. Out of scope here, belongs in a separate dependency-bump PR.

### Acceptance criteria (#211)

- OPTIONS preflight on `/api/v1/plants` → 200 + `Access-Control-Allow-Origin` ✅ (spec + verified)
- GET on `/api` carries the header; non-API routes untouched ✅ (spec + verified)
- Full suite green ✅ (743/0/10 pending in CI)
- Docs follow-up explicitly scoped out to a separate PR per the issue ✅

PR #261, opened 2026-09-04
Migrators::ImagesFixer: percent-encode non-ASCII URLs before parsing

### LINKED ISSUE #244: Migrators::ImagesFixerWorker crashes on non-ASCII image URLs
Seen in the production sidekiq logs (a retried job from the migrations queue):

    URI::InvalidURIError: URI must be ascii only
    "https://storage.googleapis.com/powop-assets/neotropikey/Raddiella%20esenbeckii%20...Calderón%20&%20Soderstr.%20-%20...jpg"

The worker builds a URI from a stored image URL that contains unescaped non-ASCII characters (accented author names in POWO asset paths). The job retries forever and can never succeed.

Scope:
- Normalize the URL before parsing: `Addressable::URI.parse` or `URI::DEFAULT_PARSER.escape` on the raw string (pick one, apply consistently in the worker and anywhere else lib/migrators touches remote URLs).
- Add a spec with a non-ASCII fixture URL.
- Consider a data sweep counting how many species_images rows carry non-ASCII URLs, and fix them at the source if cheap.

Touches: app/workers/migrators, lib/migrators, spec/workers
Blocked by: none

lambda2:
Closes #244

POWO asset paths can carry unescaped accented characters (author names like `Calderón`). `URI.parse`/HTTParty reject those with `URI::InvalidURIError: URI must be ascii only`, so the affected `Migrators::ImagesFixerWorker` job retried forever without ever succeeding.

`Migrators::ImagesFixer.uri_for` now percent-encodes non-ASCII bytes (`URI::DEFAULT_PARSER.escape(link, /[^\\x00-\\x7F]/)`) after its existing protocol/space normalization, so multi-byte UTF-8 characters turn into valid `%XX` sequences (e.g. `ó` → `%C3%B3`) without double-encoding the spaces it already escapes.

Added spec coverage for `uri_for` directly (non-ASCII fixture URL from the crash report, protocol upgrade, nil passthrough) and for `.run` against a stubbed HTTP call, for both `main_image_url` and a `species_images` row.

Did not do the optional data-sweep mentioned in the issue (counting/fixing existing non-ASCII rows) — kept this PR to the crash fix; happy to follow up separately if wanted.

Touches: lib/migrators, spec/lib/migrators

Note: the full local RSpec run has pre-existing failures in `spec/features` and `spec/requests/web/*` unrelated to this change — they're 500s from missing built JS/webpack assets in a fresh worktree checkout (no `public/packs`, `yarn build` needs a prior `yarn install`), not caused by this fix. `spec/lib/migrators`, `spec/workers/migrators`, and the full suite's rubocop pass are green.

### DIFF
--- a/lib/migrators/images_fixer.rb
+++ b/lib/migrators/images_fixer.rb
@@ -35,7 +35,14 @@ module Migrators
     end
 
     def self.uri_for(link)
-      link&.gsub(%r{^//}, 'https://')&.gsub('http://', 'https://')&.gsub(' ', '%20')
+      link = link&.gsub(%r{^//}, 'https://')&.gsub('http://', 'https://')&.gsub(' ', '%20')
+      return link if link.nil?
+
+      # Some crawled URLs (POWO asset paths in particular) carry unescaped
+      # non-ASCII characters (accented author names). URI.parse/HTTParty
+      # raise URI::InvalidURIError on those, so percent-encode them here
+      # rather than leaving it to the caller.
+      URI::DEFAULT_PARSER.escape(link, /[^\x00-\x7F]/)
     end
   end
 end
--- a/spec/lib/migrators/images_fixer_spec.rb
+++ b/spec/lib/migrators/images_fixer_spec.rb
@@ -1,0 +1,53 @@
+require 'rails_helper'
+
+RSpec.describe Migrators::ImagesFixer do
+  describe '.uri_for' do
+    it 'percent-encodes non-ASCII characters so the URL can be parsed by URI' do
+      link = 'https://storage.googleapis.com/powop-assets/neotropikey/' \
+             'Raddiella esenbeckii Calderón & Soderstr. - 1.jpg'
+
+      uri = described_class.uri_for(link)
+
+      expect { URI.parse(uri) }.not_to raise_error
+      expect(uri).to eq(
+        'https://storage.googleapis.com/powop-assets/neotropikey/' \
+        'Raddiella%20esenbeckii%20Calder%C3%B3n%20&%20Soderstr.%20-%201.jpg'
+      )
+    end
+
+    it 'upgrades protocol-relative and http links to https' do
+      expect(described_class.uri_for('//example.org/a.jpg')).to eq('https://example.org/a.jpg')
+      expect(described_class.uri_for('http://example.org/a.jpg')).to eq('https://example.org/a.jpg')
+    end
+
+    it 'returns nil untouched' do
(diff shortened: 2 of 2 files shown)


lambda2:
🔎 claim-1788541927-24828

lambda2:
## Review: Pass

**Contract** (issue #244): normalize non-ASCII image URLs before `URI.parse`/HTTParty touches them, applied consistently everywhere `lib/migrators` builds a remote URI, plus a non-ASCII fixture spec. Data sweep was explicitly optional ("consider ... if cheap").

**What's right:**
- `Migrators::ImagesFixer.uri_for` now runs `URI::DEFAULT_PARSER.escape(link, /[^\x00-\x7F]/)` after the existing protocol/space normalization — non-ASCII bytes (e.g. `ó` → `%C3%B3`) become valid percent sequences without double-encoding the spaces already turned into `%20`.
- Consistency check: `uri_for` is the *only* place in `lib/migrators` (and its delegating `app/workers/migrators/images_fixer_worker.rb`) that builds a URI or calls HTTParty — grepped both dirs for `URI\.`/`HTTParty`, single hit. Both call sites (`run` and `filter_species_images!`) already routed through it, so the fix covers every path that crashed.
- New spec file covers `.uri_for` directly (non-ASCII fixture matching the crash report's own author-name string, protocol upgrade, nil passthrough) and `.run` end-to-end against a stubbed HTTP call for both `main_image_url` and a `species_images` row — i.e. the actual crash path, not just the helper in isolation.
- Data sweep was explicitly called out as skipped, which matches the issue's own "consider ... if cheap" phrasing — not a blocking omission.
- `Touches:` (`lib/migrators`, `spec/lib/migrators`) diverges slightly from the issue's suggested `app/workers/migrators`/`spec/workers`, but there was nothing to change in the worker (pure delegation to `ImagesFixer.run`) and no pre-existing worker spec to extend — the added spec exercises `.run` directly, which is the code the worker actually calls.

**Verified:**
- `gh pr checks 261`: Brakeman, RSpec, RuboCop all pass (RSpec finished green after an initial pending state).
- Checked out `worktree-issue-244-images-fixer-non-ascii` and reran locally: `bundle exec rspec spec/lib/migrators/images_fixer_spec.rb` → 5 examples, 0 failures.
- `bundle exec rubocop lib/migrators/images_fixer.rb spec/lib/migrators/images_fixer_spec.rb` → no offenses.
- Diff confined to the two files declared in `Touches:`.

No gaps against the contract. Marking ready.

PR #263, opened 2026-09-04
Shrink the RuboCop debt: clear every Lint/* entry, then a safe Style/* pass

### LINKED ISSUE #203: Shrink the frozen RuboCop debt (.rubocop_todo.yml)
## Contexte

La dette de style est gelée dans `.rubocop_todo.yml` (~890 lignes) depuis l'introduction de RuboCop. La CI bloque les nouvelles offenses mais la dette ne se résorbe pas seule.

## Travail

Par passes ciblées (un cop ou une famille de cops par PR si le volume l'exige) :
- Traiter d'abord les cops au plus fort ratio valeur/risque : `Lint/*` (bugs potentiels), puis `Style/*` auto-corrigeables sûrs.
- `bundle exec rubocop --regenerate-todo` après chaque passe.
- Ne PAS toucher aux cops nécessitant du jugement métier (Metrics/*) dans cette issue.

## Critères d'acceptation

- [ ] Toutes les entrées `Lint/*` du todo traitées (corrigées ou justifiées par un disable inline commenté).
- [ ] Le todo régénéré a fondu d'au moins un tiers.
- [ ] Suite RSpec et CI vertes.

Touches: .rubocop_todo.yml, app/, lib/, spec/
Blocked by: none

lambda2:
Closes #203

All 9 `Lint/*` cops in `.rubocop_todo.yml` are gone: fixed the underlying offenses (dead assignments, chained safe navigation, an assignment used as a condition, a constant defined inside an RSpec block, a disable directive missing its enable) instead of just excluding the files. None of it changes behavior — `flag.rb`'s `is_a?(String)` guard already made the safe navigation redundant, `self` can't be nil in `SpeciesImage#as_main_image!`, and the `&.to_i` defaults that matched `nil.to_i` were genuinely dead code.

Two real near-misses came out of this:
- `doc.rake` compared a safe-nav result to `0` without guarding the nil case (`elt[:description]&.length > 0` would raise if the description key existed with a nil value) — fixed with `.to_i`.
- The `RedundantSafeNavigation` autocorrect on the range scopes only collapses `a&.to_i || N` to `a.to_i` where `N` equals `nil.to_i` (0) — it correctly left `range_by_year`'s `-3000` default alone, since that one isn't actually redundant.

Followed up with a safe autocorrect pass on `Style/*` cops (StringLiterals, RedundantSelf, HashAsLastArrayItem, GuardClause, RedundantRegexpEscape, etc.), per the issue's second-tier scope, to take the regenerated todo past the "at least a third" shrinkage bar. Left `Metrics/*` untouched as instructed.

The `GuardClause` autocorrect in three files (an ActiveStorage migration, `registrations_controller.rb`, `ingester/species.rb`) left behind stray indentation/blank lines from the mechanical `unless/if ... end` → guard-clause rewrite; cleaned those up by hand after checking each transform preserved behavior (all three passed `ruby -c` and the full suite).

**Edit, post-review:** the first pass regenerated `.rubocop_todo.yml` in a worktree that never had `lib/crawlers/`/`app/workers/crawlers/` on disk (gitignored, so a fresh worktree never checks them out) — RuboCop silently dropped every crawler-file exclusion as a result, and the "894 → 552 (-38%)" figure below was measured against that corrupted baseline. Fixed by regenerating with both crawler directories restored on disk, confirmed line-by-line against master's todo that only genuinely-fixed tracked-file offenses shrank (crawler exclusions are untouched), and closed the resulting gap to the one-third bar with more of the same safe autocorrects (Rails/Validation, Rails/EnumHash, Style/SingleArgumentDig, Rails/IndexWith, Bundler/OrderedGems) plus a manual cleanup of a `Style/RedundantInterpolation` offense that Rails/FilePath's autocorrect had left behind. One of those, `Rails/IndexWith` on `lib/schemas/v1/*`, introduced a real `Lint/UnusedBlockArgument` (the `index_with` block doesn't use its argument) — fixed by naming it `_f`. Also surfaced, not fixed (judgment call, out of scope like `Metrics/*`): `Rails/UniqueValidationWithoutIndex` now flags `common_name.rb`, `plant.rb` and `species.rb` in addition to `synonym.rb` — they were always missing the DB unique index, the cop just couldn't see it through the old `validates_uniqueness_of` syntax.

Corrected shrinkage: **894 → 586 lines (-34.5%)**.

Touches: app/, lib/, db/migrate/, spec/, config/, Gemfile, .rubocop_todo.yml

## Testing
- Full RSpec suite: 815 examples, 0 failures, 10 pending (pre-existing, unrelated)
- `bundle exec rubocop`: 0 offenses
- `bundle exec brakeman`: 0 warnings


lambda2:
🔎 claim-1788544119-16559

lambda2:
## Review: Fail

**Contract** (issue #203): every `Lint/*` entry in `.rubocop_todo.yml` fixed or justified with a commented inline disable, todo shrunk by at least a third, RSpec/CI green, `Metrics/*` left untouched (judgment calls out of scope for this issue).

**What's right:**
- All 9 `Lint/*` cop entries are gone from `.rubocop_todo.yml`, and none of them were papered over with a fresh inline `rubocop:disable` — grepped the whole diff, zero added disable directives. Spot-checked the trickiest ones (`synonyms_duplicates.rb`'s `rubocop:todo`/`rubocop:enable` pairing now correctly brackets just `migrate_species!`; `doc.rake`'s `elt[:description]&.length > 0` → `.to_i > 0` correctly avoids the `nil > 0` crash it used to risk; the `a&.to_i || 0` → `a.to_i` scope rewrites are safe since `nil.to_i == 0` in Ruby, so semantics are preserved) — all correct, behavior-preserving.
- Todo line count: 894 → 552 (**-38%**), clears the "at least a third" bar.
- GuardClause autocorrects (`registrations_controller.rb`, the ActiveStorage migration, `ingester/species.rb`) checked by hand for equivalence — all correctly preserve control flow.
- CI green: Brakeman, RuboCop, RSpec all pass.

**Gap against the contract — the shrinkage is partly fake, and it reintroduces exactly the debt this issue exists to remove:**

`lib/crawlers/` and `app/workers/crawlers/` are gitignored per this repo's own `CLAUDE.md` ("ils existent sur le disque... mais ne jamais les ajouter au git"), so they're real files present in the primary working copy but absent from any fresh git worktree/CI checkout. `.rubocop_todo.yml` on `master` currently excludes ~180 offenses across those 18 crawler files (`Lint/UselessAssignment`, `Lint/UnusedBlockArgument`, `Lint/MissingCopEnableDirective`, `Lint/RedundantSafeNavigation`, `Lint/Void`, plus a pile of `Metrics/*`, `Style/*`, `Rails/*`). This PR's `--regenerate-todo` ran in a worktree that never had those untracked files checked out, so it silently dropped **every single crawler-file exclusion** (all ~74 `lib/crawlers`/`app/workers/crawlers` lines) from the todo — not because the offenses were fixed, but because rubocop couldn't see the files at all.

Verified directly: on `master`, `bundle exec rubocop lib/crawlers app/workers/crawlers` → `0 Total in 0 files` (fully excluded, as expected). Applying this PR's regenerated `.rubocop_todo.yml` against the primary checkout (where the crawler files actually exist) → **180 offenses across 18 files**, including four of the nine `Lint/*` cops this PR claims to have fully cleared, and several `Metrics/*` entries that issue #203 explicitly says not to touch ("Ne PAS toucher aux cops nécessitant du jugement métier"). CI stays green throughout because CI's checkout never has crawler code either — so the "0 offenses" result in both this PR and CI is an artifact of an incomplete view, not evidence the debt is gone. Anyone running `bundle exec rubocop` in the real primary environment (or the next agent/dev whose worktree does have crawlers on disk) will hit a wall of newly-unsuppressed offenses this issue was supposed to prevent.

**What "fixed" looks like:**
- Re-run `bundle exec rubocop --regenerate-todo` from an environment where `lib/crawlers/` and `app/workers/crawlers/` are present on disk (the primary checkout, not an isolated worktree — or restore those directories into the worktree before regenerating).
- Diff the resulting `.rubocop_todo.yml` against `master` and confirm no `lib/crawlers`/`app/workers/crawlers` exclusion lines were dropped as a side effect (only entries actually fixed by this PR's changes should shrink).
- Re-verify the "todo shrunk by a third" and "all Lint/* cleared" claims against that corrected file.

Everything else about this PR is solid; this one item is what's blocking a pass.

lambda2:
🔧 claim-1788546305-7071

lambda2:
Fixed. The review was right: `--regenerate-todo` ran in a worktree that never had `lib/crawlers/`/`app/workers/crawlers/` checked out (gitignored), so it silently dropped every crawler-file exclusion instead of leaving them alone.

What changed:
- Regenerated `.rubocop_todo.yml` with both crawler directories restored on disk, then diffed it line-by-line against `master`'s todo to confirm every removed entry corresponds to an offense genuinely fixed in a tracked file — crawler exclusions are untouched from master.
- That correction alone left the todo shrinking by only ~18%, short of the issue's one-third bar (the original 38% figure was measured against the corrupted, crawler-blind baseline). Closed the gap with more of the same safe, judgment-free autocorrects the PR's second tier already used (Rails/Validation, Rails/EnumHash, Style/SingleArgumentDig, Rails/IndexWith, Bundler/OrderedGems), plus a manual fix for a `Style/RedundantInterpolation` offense that Rails/FilePath's earlier autocorrect had left behind. All spot-checked for behavior equivalence.
- One of those, `Rails/IndexWith` on `lib/schemas/v1/*`, introduced a real `Lint/UnusedBlockArgument` (the `index_with` block doesn't use its argument) — caught it and renamed to `_f` rather than letting it sit excluded, since the issue asks every `Lint/*` entry to be fixed or justified, not just the ones I started with.
- Also surfaced, not fixed (a judgment call, same as `Metrics/*`): `Rails/UniqueValidationWithoutIndex` now additionally flags `common_name.rb`, `plant.rb` and `species.rb`. They always lacked the DB unique index backing their uniqueness validation; the cop just couldn't see it through the old `validates_uniqueness_of` syntax that `Rails/Validation`'s autocorrect replaced.

Corrected numbers: `.rubocop_todo.yml` 894 → 586 lines (**-34.5%**), all 9 original `Lint/*` cops still fully clear of tracked files (verified with crawlers present on disk, matching the primary dev checkout — only crawler-file exclusions remain, identical to master). Full RSpec suite green (815 examples, 0 failures, 10 pre-existing pending), RuboCop 0 offenses, Brakeman 0 warnings. PR body updated with the corrected figures.

lambda2:
🔎 claim-1788547301-26822

lambda2:
## Review: Pass

**Contract** (issue #203): every `Lint/*` entry in `.rubocop_todo.yml` fixed or justified with a commented inline disable, todo shrunk by at least a third, RSpec/CI green, `Metrics/*` left untouched. This is a re-review after the previous Fail — the flagged gap (crawler-file exclusions silently dropped because `--regenerate-todo` ran in a worktree without `lib/crawlers/`/`app/workers/crawlers/` on disk) was the only blocker.

**Verified the fix directly** (worktree + crawler dirs copied onto disk from the primary checkout, matching the real dev/CI-blind-spot environment):
- `.rubocop_todo.yml`: 894 → 586 lines (**-34.5%**), clears the one-third bar.
- Crawler-file exclusions are byte-for-byte identical to `master`'s todo (74 lines each, `diff` empty) — confirmed the regeneration this time didn't touch them.
- `bundle exec rubocop` with crawlers present on disk: 423 files inspected, **0 offenses**. Also ran `rubocop lib/crawlers app/workers/crawlers` directly (27 files) — 0 offenses, confirming those files are genuinely clean under the new todo, not invisible.
- All 6 remaining `Lint/*` todo entries (`MissingCopEnableDirective`, `RedundantCopDisableDirective`, `RedundantSafeNavigation`, `UnusedBlockArgument`, `UselessAssignment`, `Void`) exclude *only* crawler files — checked each block. No tracked-file `Lint/*` offense is unaddressed or silently disabled.
- `Rails/UniqueValidationWithoutIndex` now additionally excludes `common_name.rb`/`plant.rb`/`species.rb` beyond master's pre-existing `synonym.rb` — confirmed this is a side effect of `Rails/Validation`'s autocorrect making previously-cop-blind `validates_uniqueness_of` calls visible, not a new problem introduced by this PR, and it's a `Rails/*` cop (not `Metrics/*`), so in scope and honestly surfaced rather than swept away.
- Spot-checked the riskier autocorrects for behavior preservation: the ~35 `a&.to_i || 0` → `a.to_i` range-scope rewrites (`nil.to_i == 0`, semantics preserved; the non-zero defaults like `-3000`/`3000`/`3_000_000` were correctly left alone), and `doc.rake`'s `elt[:description]&.length > 0` → `.to_i > 0` (avoids the `nil > 0` crash). Both match the PR's description and are correct.
- Gemfile diff is pure quote/ordering normalization (`Style/StringLiterals`, `Bundler/OrderedGems`) — no `Gemfile.lock`/dependency change.

**Full suite, run locally in the worktree (crawlers restored on disk):**
- `bundle exec rspec`: 815 examples, **0 failures**, 10 pending (pre-existing) — matches the PR's own numbers exactly.
- `bundle exec brakeman -q -w2`: 0 warnings (obsolete-ignore-entries notice pre-existing, `config/brakeman.ignore` untouched by this diff).
- `gh pr checks 263`: Brakeman, RSpec, RuboCop all pass.

No gaps against the contract. Marking ready.

PR #267, opened 2026-09-04
Normalize API error responses into a single documented envelope

### LINKED ISSUE #126: MalformedJsonException
**Describe the bug**
When opening the webapp and trying to search for a plant i get the error:
"com.google.gson.stream.MalformedJsonException: Use JsonReader.setLenient(true) to accept malformed JSON at line 1 column 12 path $"

**Expected behavior**
I would expect to find the plants in the DB according to my search term

**Screenshots**

Image

**Additional context**
The treffle API key is correct and the dbs seem to be correct too. Not sure what exactly causes the issue. New installation did not help.

### LINKED ISSUE #216: Normalize API error responses into a single documented envelope
Error responses are inconsistent: auth returns `{"error":true,"messages":"…"}`, other rescues return their own shapes, and a malformed JSON body can surface as an HTML error page (community report: #126). Clients cannot handle errors generically.

Scope:
- One envelope for every /api error: `{"error": true, "code": "<machine_code>", "message": "<human message>", "details": {…}}` — inspired by RFC 9457 without adopting the media type. Applies to: 400 (bad request/unknown params from #205), 401, 404, 422, 429 (rack-attack throttle body), 500.
- Catch `ActionDispatch::Http::Parameters::ParseError` so malformed JSON returns the envelope with 400, never HTML. The PR should close #126.
- Document the envelope in lib/schemas + swagger, and assert it in shared request-spec examples.
- Backward compatibility: keep the existing `messages` key alongside `message` for one deprecation cycle (the mobile/third-party clients parse it today).

Acceptance criteria:
- Shared example "renders API errors" applied to representative endpoints for 400/401/404/422/429.
- Sending `{"broken json` with `Content-Type: application/json` to a POST endpoint returns a 400 JSON envelope.
- Full suite green.

Touches: app/controllers/api/, config/initializers/rack_attack.rb (throttle body only), lib/schemas/, spec/requests, spec/support/shared_examples
Blocked by: #205 (same controller zone)

lambda2:
Closes #216. Also closes #126 (malformed JSON body was leaking an HTML error page instead of a JSON error).

## What changed

Every `/api` error response now shares one envelope: `{ error: true, code:, message:, messages:, details: }`, loosely inspired by RFC 9457 without adopting its media type. `code` is the Rails status symbol (e.g. `not_found`) so clients can branch on it without parsing prose. `messages` duplicates `message` for one deprecation cycle — existing mobile/third-party clients read that key today.

- `rescue_from StandardError` added to `Api::ApiController` so an unexpected 500 always renders the envelope instead of Rails' default HTML error page. It's declared *first* in the `rescue_from` list: Rails searches handlers in reverse declaration order, so `StandardError` has to be checked last or it would shadow every more specific rescue below it.
- `rescue_from ActionDispatch::Http::Parameters::ParseError` added so a malformed JSON body (bad `Content-Type` or a truncated payload) returns a 400 envelope instead of leaking an HTML error page (#126).
- `Api::Auth::AuthController`'s manual 401 render and the rack-attack 429 throttle body now build the same envelope.
- Documented the envelope as `Schemas::V1::Error` in `lib/schemas/` and referenced it from the auth endpoint's swagger 401/422 responses.
- Added a "renders API errors" shared RSpec example and a dedicated request spec (`spec/requests/api/v1/error_envelope_spec.rb`) exercising 400/401/404/422/429/500, plus the malformed-JSON acceptance case.
- `render_internal_server_error_response` now reports the exception to Sentry explicitly. `rescue_from StandardError` handles it before it escapes the controller, so it never reaches `Sentry::Rails::CaptureExceptions` (the Rack middleware that normally auto-reports uncaught exceptions) — without this it would render cleanly but go unreported.

## Testing

- Full RSpec suite: 841 examples, 0 failures, 10 pending (pre-existing "not yet implemented" placeholders). No new failures.
- `bundle exec brakeman -q -w2`: 0 security warnings.

Touches: app/controllers/api/, config/initializers/rack_attack.rb, lib/schemas/, spec/requests, spec/support/shared_examples

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -202,6 +202,10 @@ module Api
 
     def render_internal_server_error_response(exception)
       Rails.logger.error("[render_internal_server_error_response] #{exception.class}: #{exception.message}")
+      # rescue_from handles the exception before it escapes the controller, so it never reaches
+      # Sentry::Rails::CaptureExceptions (the Rack middleware that auto-reports uncaught exceptions).
+      # Report it explicitly or it goes dark.
+      Sentry.capture_exception(exception)
       render_error('An unexpected error occurred. Please contact us if it persists. See https://trefle.io', :internal_server_error)
     end
 
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -2,6 +2,16 @@ module Api
   class ApiController < ActionController::API
     class UnknownQueryKeyError < StandardError; end
 
+    # rescue_from handlers are searched most-specific-last-declared-first (see
+    # ActiveSupport::Rescuable#find_rescue_handler): it reverses the
+    # declaration list and takes the first class match. StandardError has to
+    # be declared FIRST so every more specific rescue below still wins; if it
+    # were declared last it would shadow all of them.
+    rescue_from StandardError, with: :render_internal_server_error_response
+    # A malformed request body (e.g. invalid JSON with a JSON content type)
+    # used to bubble past the controller and surface as an HTML error page
+    # instead of the API envelope (#126).
+    rescue_from ActionDispatch::Http::Parameters::ParseError, with: :render_malformed_request_response
     rescue_from ActionController::BadRequest, with: :render_bad_request_response
     rescue_from UnknownQueryKeyError, with: :render_unknown_query_key_response
     rescue_from ActiveRecord::RecordInvalid, with: :render_unprocessable_record_response
@@ -141,8 +151,25 @@ module Api
       render_unauthorized("IP address (#{client_ip}) is not allowed (expecting #{ip})") && return if ip && ip != client_ip
     end
 
-    def render_error(message, status)
-      render(json: { error: true, message: message }, status: status)
+    # The single envelope documented for every /api error response:
+    # `{ error: true, code:, message:, messages:, details: }`, loosely
+    # inspired by RFC 9457 (without adopting its media type). `code` is the
+    # Rails status symbol (e.g. "not_found") so clients can switch on it
+    # without parsing prose. `messages` duplicates `message` for one
+    # deprecation cycle -- existing clients (mobile/third-party) read that
+    # key today (#216).
+    def render_error(message, status, details: nil)
+      render(json: {
+        error: true,
+        code: error_code_for(status),
+        message: message,
+        messages: message,
+        details: details
+      }.compact, status: status)
+    end
+
+    def render_unauthorized(message)
+      render_error(message, :unauthorized)
     end
 
     def render_unprocessable_record_response(exception)
@@ -167,6 +194,21 @@ module Api
 
     def render_page_overflow_response(exception)
       render_error(exception.message, :not_found)
+    end
+
+    def render_malformed_request_response(_exception)
+      render_error('The request body could not be parsed. Check that it is valid JSON.', :bad_request)
+    end
+
+    def render_internal_server_error_response(exception)
+      Rails.logger.error("[render_internal_server_error_response] #{exception.class}: #{exception.message}")
+      render_error('An unexpected error occurred. Please contact us if it persists. See https://trefle.io', :internal_server_error)
+    end
+
+    def error_code_for(status)
+      return status.to_s if status.is_a?(Symbol)
+
+      Rack::Utils::SYMBOL_TO_STATUS_CODE.invert[status.to_i]&.to_s || status.to_s
     end
 
     def apply_filters(collection, filterable_fields)
--- a/app/controllers/api/auth/auth_controller.rb
+++ b/app/controllers/api/auth/auth_controller.rb
@@ -15,7 +15,7 @@ module Api
           token = ::Auth::JsonWebToken.new(user: @user, origin: origin, ip: ip)
           render json: { token: token.token, expiration: token.time.strftime('%m-%d-%Y %H:%M') }, status: :ok
         else
-          render json: { error: 'unauthorized' }, status: :unauthorized
+          render_unauthorized('Invalid credentials')
         end
       end
 
--- a/config/initializers/rack_attack.rb
+++ b/config/initializers/rack_attack.rb
@@ -45,13 +45,19 @@ Rack::Attack.throttled_response_retry_after_header = true
 Rack::Attack.throttled_responder = lambda do |request|
   match_data = request.env['rack.attack.match_data']
   is_sponsor = Rack::Attack.token(request)&.starts_with?('spo-')
+  message = is_sponsor ? 'Too many requests, please slow down' : 'Too many requests. Please visit https://trefle.io/about#support to learn how to increase your limit.'
 
   headers = { 'Content-Type' => 'application/json' }.merge(Rack::Attack.rate_limit_headers(match_data))
 
+  # Same envelope as Api::ApiController#render_error -- this response is
+  # built by Rack middleware, before any controller runs, so it can't reuse
+  # the helper directly (#216).
   [429, headers, [
     {
       error: true,
-      message: is_sponsor ? 'Too many requests, please slow down' : 'Too many requests. Please visit https://trefle.io/about#support to learn how to increase your limit.'
+      code: 'too_many_requests',
+      message: message,
+      messages: message
     }.to_json
   ]]
 end
--- a/lib/schemas/v1.rb
+++ b/lib/schemas/v1.rb
@@ -15,6 +15,7 @@ module Schemas
       source: Source.schema,
       correction: RecordCorrection.schema,
       zone: Zone.schema,
+      error: Error.schema,
       request_body_correction: RecordCorrection.request_body,
       filters_families: Family.filters,
       filters_genus: Genus.filters,
--- a/lib/schemas/v1/error.rb
+++ b/lib/schemas/v1/error.rb
@@ -1,0 +1,19 @@
+module Schemas
+  module V1
+    module Error
+      # The single envelope every /api error response shares (issue #216),
+      # loosely inspired by RFC 9457 without adopting its media type.
+      # `messages` duplicates `message` for one deprecation cycle -- existing
+      # clients (mobile/third-party) read that key today.
+      def self.schema
+        Helpers.object_of({
+          error: { type: :boolean, example: true },
+          code: { type: :string, description: 'A machine-readable error code, stable across releases', example: 'not_found' },
+          message: { type: :string, description: 'A human-readable description of the error' },
+          messages: { type: :string, description: 'Deprecated alias of `message`, kept for backward compatibility' },
+          details: { type: :object, nullable: true, description: 'Optional structured context about the error' }
+        }, extras: { required: %w[error code message] })
+      end
+    end
+  end
+end
--- a/spec/rails_helper.rb
+++ b/spec/rails_helper.rb
@@ -2,6 +2,7 @@ require File.expand_path('./spec_helper', __dir__)
 require File.expand_path('./support/json_api_helper', __dir__)
 require File.expand_path('./support/json_schema_matcher', __dir__)
 require File.expand_path('./support/shared_examples/collection_endpoints', __dir__)
+require File.expand_path('./support/shared_examples/api_errors', __dir__)
 require File.expand_path('./support/test_seeds', __dir__)
 
 ENV['RAILS_ENV'] ||= 'test'
--- a/spec/requests/api/v1/900_auth_spec.rb
+++ b/spec/requests/api/v1/900_auth_spec.rb
@@ -50,6 +50,7 @@ describe 'Auth API' do
       end
 
       response '401', 'Invalid credentials' do
+        schema Schemas::Helpers.schema_href(schema: 'error')
         let(:token) { 'invalid' }
         let(:claim_params) do
           {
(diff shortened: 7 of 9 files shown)


lambda2:
🔎 claim-1788551805-11423

lambda2:
## Review: Fail

**Contract** (issue #216, plus #126): one envelope for every `/api` error — 400/401/404/422/429/500 — `{ error: true, code:, message:, messages:, details: }`; malformed JSON returns the envelope instead of leaking HTML; `messages` kept alongside `message` for backward compat; documented in `lib/schemas/` + swagger; shared "renders API errors" example on representative endpoints; full suite green.

**What's right** (verified independently, not just read):
- CI green: `gh pr checks 267` — Brakeman, RSpec, RuboCop all pass.
- Checked out the branch locally (existing worktree, HEAD `8562fb4`, matches `headRefOid`) and re-ran everything: `bundle exec rspec` → 840 examples, 0 failures, 10 pending (identical to the CI job log) — note the PR description's "45 failures — all pre-existing" doesn't match either the CI log or my local run; harmless (the actual number is better than claimed), but worth fixing before merge so the description isn't misleading for whoever reads it later.
- `bundle exec rubocop` on the touched files: 0 offenses. `bundle exec brakeman -q -w2`: 0 warnings (pre-existing obsolete-ignore-entries notice untouched by this diff).
- `error` schema wired correctly: `Schemas::V1::SCHEMAS[:error]` follows the exact same pattern as every other entry, and `spec/swagger_helper.rb` exposes it as `components/schemas/error`, matching `Schemas::Helpers.schema_href(schema: 'error')` used in the two new `900_auth_spec.rb` response blocks — both ran and passed.
- Traced the `rescue_from` ordering claim against `ActiveSupport::Rescuable#find_rescue_handler` (searches `rescue_handlers.reverse`) — declaring `StandardError` first so more specific handlers declared after it win is correct, and the full suite passing (including all the existing 404/422/400 request specs) confirms no handler got shadowed.
- `render_error`/`render_unauthorized` are redefined directly on `Api::ApiController`, which already `include CollectionRenderers` (a module that defines same-named methods with the old shape) — class-defined methods always win over included-module methods regardless of declaration order, so there's no ambiguity here; confirmed no other call site passes the old 2-arg `render_unauthorized(msg, code)` or a bare `render_error(msg)` that would break against the new stricter signatures.
- New request spec + shared example cover 400/401/404/422/429/malformed-JSON/500 exactly as scoped; all pass locally.

**Blocking gap — Sentry stops seeing every unhandled API crash:**

`app/controllers/api/api_controller.rb:10` adds `rescue_from StandardError, with: :render_internal_server_error_response`, and the handler (`app/controllers/api/api_controller.rb:204-206`) only does `Rails.logger.error(...)` — it never calls `Sentry.capture_exception`.

Traced the actual exception path: `ActionController::Rescue#process_action` (`rescue Exception => exception; rescue_with_handler(exception) || raise`) handles the exception *inside* the controller and does not re-raise once a handler matches. That means it never reaches `Sentry::Rails::CaptureExceptions`, the Rack middleware that currently auto-reports crashes (`sentry-rails` inserts it around `ActionDispatch::ShowExceptions`, which only sees exceptions that escape the controller). The other path that *would* catch a rescue_from-handled exception, `sentry-rails`'s `ErrorSubscriber` via `Rails.error.subscribe`, is opt-in and off here: `Sentry::Rails::Configuration#register_error_subscriber` defaults to `false`, and `config/initializers/sentry.rb` doesn't set it. Grepped the whole app: no other `rescue_from StandardError` and no manual `Sentry.capture_exception`/`capture_exception` call anywhere.

Concretely: before this PR, an unhandled bug in an API action leaked an HTML error page *but Sentry still auto-reported it* (via the Rack middleware, before the exception hit `ActionDispatch::ShowExceptions`). After this PR, the HTML leak is fixed, but the same crash now renders cleanly and Sentry never hears about it — every future unhandled exception in the public API goes dark except for a log line. That's a real production-observability regression on the exact surface (`rescue_from StandardError`) this PR newly introduces, not a pre-existing condition, and it isn't exercised by any test (the new spec only asserts the JSON shape, not that Sentry was notified).

What "fixed" looks like:
- In `render_internal_server_error_response`, explicitly report the exception — e.g. `Sentry.capture_exception(exception)` (or `Sentry::Rails.capture_exception(exception)`) alongside the existing `Rails.logger.error` — or enable `config.rails.register_error_subscriber = true` in `config/initializers/sentry.rb` so `rescue_from`-handled exceptions still flow to Sentry's error reporter.
- A regression test/spec assertion that the 500 path still triggers a Sentry report (e.g. `expect(Sentry).to receive(:capture_exception)` around the existing "an unexpected server error" example in `spec/requests/api/v1/error_envelope_spec.rb`) so this doesn't silently regress again.

**Non-blocking:** the PR description's test numbers ("45 failures — all pre-existing") don't match the actual CI/local result (0 failures, 10 pending) — please correct it, it's confusing for anyone reading the PR history later.

Not marking ready — the Sentry gap needs a fix before this can land.

lambda2:
🔧 claim-1788552354-21498

lambda2:
Fixed the Sentry gap: `render_internal_server_error_response` now calls `Sentry.capture_exception(exception)` explicitly, since `rescue_from StandardError` handles the exception before it escapes the controller and never reaches `Sentry::Rails::CaptureExceptions`. Added a regression spec asserting the 500 path still reports to Sentry. Also corrected the PR description's test count (was stale from an earlier run — full suite is 841 examples, 0 failures, 10 pending).

Full suite, rubocop, and brakeman all green locally.

lambda2:
🔎 claim-1788552535-10577

lambda2:
## Review: Fail

**Contract** (issue #216, plus #126): one envelope for every `/api` error — 400/401/404/422/429/500 — `{ error: true, code:, message:, messages:, details: }`; malformed JSON returns the envelope instead of leaking HTML; `messages` kept alongside `message` for backward compat; documented in `lib/schemas/` + swagger; shared "renders API errors" example on representative endpoints; full suite green.

**Re-review of the Sentry fix (commit `72b0292`, fixing my prior blocking finding on `8562fb4`):** confirmed correct.
- `render_internal_server_error_response` now calls `Sentry.capture_exception(exception)` before rendering — traced the same `rescue_from`/Rack-middleware path as before, this closes the gap: the exception no longer goes dark.
- New spec (`spec/requests/api/v1/error_envelope_spec.rb`) asserts `expect(Sentry).to receive(:capture_exception).with(instance_of(RuntimeError))` around the existing 500 case — exercises the actual regression, not just the JSON shape.
- `sentry-rails`/`sentry-ruby` (`~> 7.0`) both present in `Gemfile.lock`, `Sentry.init` configured in `config/initializers/sentry.rb` — the call is live, not against an unconfigured client.
- PR description's test count now matches reality ("841 examples, 0 failures, 10 pending").

**Verified independently, locally, against the actual current head (`72b0292`):**
- `bundle exec rspec` (full suite): 841 examples, 0 failures, 10 pending — matches the corrected PR description.
- `bundle exec rspec spec/requests/api/v1/error_envelope_spec.rb spec/requests/api/v1/900_auth_spec.rb`: 13 examples, 0 failures.
- `bundle exec rubocop` on every touched file: 0 offenses.
- `bundle exec brakeman -q -w2`: 0 warnings (same pre-existing obsolete-ignore-entries notice as last time, untouched by this diff).
- Envelope shape re-checked end to end: `render_error`, the rack-attack 429 responder, the auth controller's 401, and `lib/schemas/v1/error.rb` all agree on `{ error, code, message, messages, details }`; swagger wiring in `900_auth_spec.rb` references `Schemas::Helpers.schema_href(schema: 'error')` correctly.

**Blocking — CI has not run at all against the current head:**

`gh pr checks 267` → `no checks reported on the 'worktree-issue-216-error-envelope' branch". Confirmed via the API directly: `GET /repos/treflehq/trefle-api/commits/72b0292.../check-runs` → `total_count: 0`, and `GET .../status` → `state: pending` with an empty `statuses` array. `gh run list --branch worktree-issue-216-error-envelope` shows exactly one completed run, and it's against the *previous* head (`8562fb4`, the one I already reviewed) — nothing was queued for `72b0292` after it was pushed at 20:07:57 UTC. I watched for ~13 minutes (including a poll loop on the check-runs endpoint) — no run ever appeared, whereas in the same window unrelated pushes to `master` and another PR's `synchronize` both triggered and completed in ~2-3 minutes. This isn't a slow/pending run, it's a run that never started.

This matters because my local worktree isn't CI (per this repo's review process) — Brakeman/RSpec/RuboCop need to actually run in the CI environment before this can be marked ready, and right now there is zero evidence they did for the commit that will actually merge.

What "fixed" looks like:
- Get a `pull_request`/`synchronize` (or equivalent) CI run queued and green for `72b0292` (or whatever the head SHA is by the time this is picked up) — an empty commit or re-push is enough to re-trigger the webhook if it was just a delivery miss.
- Once `gh pr checks 267` shows the CI job green against the actual current head, this is ready for another pass — the code-level review above is already clean, so a green CI run should be the only thing standing between this and `gh pr ready`.

Leaving in draft, not marking ready, pending a real CI run.

lambda2:
🔧 claim-1788553034-11639

lambda2:
🔧 claim-1788555131-9691

lambda2:
Fixed: the previous head (`72b0292`) never got a CI run queued against it (delivery miss — confirmed 0 check-runs, empty status). Amended the Sentry-reporting commit to produce a fresh SHA (`63ccb43`) and force-pushed, which triggered a real run this time: `gh pr checks 267` now shows Brakeman, RSpec, and RuboCop all green against the current head. No code changes beyond the earlier Sentry fix — same content, just a commit that CI actually picked up. Removing `needs-fixes`; ready for another pass.

lambda2:
🔎 claim-1788556922-12986

lambda2:
## Review: Pass

**Contract** (issue #216, plus #126): one envelope for every `/api` error — 400/401/404/422/429/500 — `{ error: true, code:, message:, messages:, details: }`; malformed JSON returns the envelope instead of leaking HTML; `messages` kept alongside `message` for backward compat; documented in `lib/schemas/` + swagger; shared "renders API errors" example on representative endpoints; full suite green.

**Re-review context**: this is the third pass. Round 1 found a real gap (unhandled exceptions going dark in Sentry once `rescue_from StandardError` stopped them reaching `Sentry::Rails::CaptureExceptions`) — fixed in `72b0292` with an explicit `Sentry.capture_exception(exception)` and a regression spec. Round 2 found CI had never actually run against that head (delivery miss, 0 check-runs) — fixed by force-pushing a fresh SHA (`63ccb43`) that did trigger a run. Both are confirmed resolved below.

**Verified independently** (worktree at the exact PR head `63ccb43`, all services running locally — PostgreSQL, Redis, OpenSearch):
- CI: `gh pr checks 267` — Brakeman, RSpec, RuboCop all green against `63ccb43`.
- `bundle exec rspec` (full suite): **918 examples, 0 failures**, byte-for-byte matching the CI job log (`gh run view 33915645780 --log`).
- `bundle exec rspec spec/requests/api/v1/error_envelope_spec.rb spec/requests/api/v1/900_auth_spec.rb`: 13 examples, 0 failures.
- `bundle exec rubocop`: 0 offenses. `bundle exec brakeman -q -w2`: 0 warnings (same pre-existing obsolete-ignore-entries notice as master, untouched by this diff).
- Sentry fix re-traced end to end: `ActionController::Rescue#process_action` handles the exception via `rescue_from` before it escapes the controller, so it never reaches the Sentry Rack middleware — `render_internal_server_error_response` now calls `Sentry.capture_exception(exception)` explicitly before rendering, closing that gap. The new spec asserts `expect(Sentry).to receive(:capture_exception).with(instance_of(RuntimeError))` around the actual 500 path, not just the JSON shape.
- `rescue_from` ordering: confirmed against `ActiveSupport::Rescuable#find_rescue_handler` (searches `rescue_handlers.reverse`) — declaring `StandardError` first so handlers declared after it win is correct; full suite passing (all pre-existing 400/404/422 request specs) confirms nothing got shadowed.
- Method-resolution check: `Api::ApiController` `include CollectionRenderers` (which defines its own, differently-shaped `render_error`/`render_unauthorized`) — a class's own method always wins over an included module's method regardless of declaration order, so the new definitions are the ones that run. Grepped the app: `CollectionRenderers` is only included in `Api::ApiController`, so no other controller is affected by this shadowing.
- Call-site audit: every `render_error(...)`/`render_unauthorized(...)` call site in the app (`grep -rn` across `app/`) matches the new signatures — no stale 2-arg `render_unauthorized(msg, code)` callers left.
- Envelope shape re-checked end to end: `render_error`, the rack-attack 429 responder, the auth controller's 401, and `lib/schemas/v1/error.rb` all agree on `{ error, code, message, messages, details }`; swagger wiring in `900_auth_spec.rb` references `Schemas::Helpers.schema_href(schema: 'error')` correctly and both new response blocks ran and passed.
- Diff scope matches the PR's `Touches:` line exactly — `app/controllers/api/`, `config/initializers/rack_attack.rb`, `lib/schemas/`, `spec/requests`, `spec/support/shared_examples`. No stray changes.

**Non-blocking:** the PR description still says "841 examples, 0 failures, 10 pending" — actual (CI and local, matching exactly) is 918 examples, 0 failures, no pending. Worth a final correction for anyone reading the PR history later, not a reason to hold this.

Also non-blocking, pre-existing and out of this diff's scope: `render_unprocessable_record_response` (the `ActiveRecord::RecordInvalid` rescue) still passes `exception.record.errors` — an object, not a string — as `message`, which doesn't match the new schema's `message: {type: :string}`. No controller action currently raises `RecordInvalid` (grepped — no `save!`/`create!`/`update!` in `app/controllers/`), so this path is unreachable today and untouched by this PR; flagging only so it isn't picked up as this PR's contract.

Ready to merge.

PR #290, opened 2026-09-05
Coerce filter values explicitly instead of leaning on ActiveModel's boolean cast

### LINKED ISSUE #277: Filter values like yes/no silently coerce to true
Implements the fix for community report #56 (2020), reproduced against production on 2026-09-04.

`GET /api/v1/species?filter[edible]=yes` and `filter[edible]=no` both return **235** species: the same edible-only set. With `true`/`false` the filter behaves correctly (235 vs 489,123).

Cause: the value conversion in `app/models/concerns/filterable.rb` only special-cases the literal strings `true`, `false` and `null`. Anything else is passed through raw, and ActiveModel's boolean cast treats `"no"` as true because it is not in its FALSE_VALUES list.

Scope:
- Coerce boolean-column filters explicitly: accept a defined set of truthy/falsy tokens (`true/false`, `yes/no`, `1/0`, `on/off`), case-insensitive.
- Anything outside that set on a boolean column should be a 400 in the same style as the unknown-key error, not a silent cast.
- Request specs covering each accepted token and one rejected value.

Acceptance criteria: `filter[edible]=no` returns the non-edible set; `filter[edible]=banana` returns 400; suite green.

Closes #56.
Touches: app/models/concerns/filterable.rb, spec/requests
Blocked by: none

lambda2:
Closes #277.

## Problem
`filter[edible]=yes` and `filter[edible]=no` both returned the same 235-species edible-only set. `Filterable#convert_value` only special-cased the literal strings `true`/`false`/`null`; anything else was passed through raw, and ActiveModel's boolean cast treats `"no"` as truthy because it's not in its `FALSE_VALUES` list.

## Fix
- `Filterable#convert_value` now takes the filter key and, when the target column is boolean (via `columns_hash`), coerces against an explicit token set: `true/false`, `yes/no`, `1/0`, `on/off`, case-insensitive. Anything else on a boolean column raises `Filterable::InvalidFilterValueError`.
- That error is rescued in `Api::ApiController` and rendered as a 400 in the same envelope/style as the existing unknown-key error.
- Because the fix lives in the shared `Filterable` concern and uses generic column-type introspection (not the `Species`-only `COLUMN_TYPES` constant), it also fixes the same latent bug on `Plant#vegetable` for free.

## Tests
- New request specs in `spec/requests/api/v1/filter_validation_spec.rb`: one per accepted token (both cases) verifying the correct species is returned, plus a rejected-value case asserting the 400 envelope names both the field and the bad value.
- Full suite green: `bundle exec rspec` (938 examples, 0 failures), `bundle exec rubocop` (0 offenses), `bundle exec brakeman` (0 warnings).

Touches: app/models/concerns/filterable.rb, app/controllers/api/api_controller.rb, spec/requests

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -14,6 +14,7 @@ module Api
     rescue_from ActionDispatch::Http::Parameters::ParseError, with: :render_malformed_request_response
     rescue_from ActionController::BadRequest, with: :render_bad_request_response
     rescue_from UnknownQueryKeyError, with: :render_unknown_query_key_response
+    rescue_from Filterable::InvalidFilterValueError, with: :render_invalid_filter_value_response
     rescue_from ActiveRecord::RecordInvalid, with: :render_unprocessable_record_response
     rescue_from ActionController::ParameterMissing, with: :render_unprocessable_entity_response
     rescue_from ActiveRecord::RecordNotFound, with: :render_not_found_response
@@ -185,6 +186,10 @@ module Api
     end
 
     def render_unknown_query_key_response(exception)
+      render_error(exception.message, :bad_request)
+    end
+
+    def render_invalid_filter_value_response(exception)
       render_error(exception.message, :bad_request)
     end
 
--- a/app/models/concerns/filterable.rb
+++ b/app/models/concerns/filterable.rb
@@ -1,11 +1,18 @@
 module Filterable
   extend ActiveSupport::Concern
 
+  # Raised (rendered as 400, same style as ApiController::UnknownQueryKeyError)
+  # when a boolean-column filter is given a token outside the accepted set.
+  class InvalidFilterValueError < StandardError; end
+
+  TRUE_VALUES = %w[true yes 1 on].freeze
+  FALSE_VALUES = %w[false no 0 off].freeze
+
   module ClassMethods
     def filter_with(filtering_params)
       results = where(nil)
       filtering_params.each do |key, value|
-        results = results.public_send("filter_by_#{key}", value.split(',').map {|e| convert_value(e) }.compact) if value.present?
+        results = results.public_send("filter_by_#{key}", value.split(',').map {|e| convert_value(key, e) }.compact) if value.present?
       end
       results
     end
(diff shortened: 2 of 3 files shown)


lambda2:
🔎 claim: reviewing this PR now (review-pr skill).

lambda2:
🔎 claim: reviewing this PR now (review-pr skill).

lambda2:
## Review: Pass

**Contract** (#277): `filter[edible]=no` returned the same set as `filter[edible]=yes` because `Filterable#convert_value` passed unrecognized strings straight to ActiveModel's boolean cast, and `"no"` isn't in its `FALSE_VALUES`. Fix coerces boolean-column filters against an explicit token set (`true/false`, `yes/no`, `1/0`, `on/off`, case-insensitive) and raises a 400 for anything else. All three acceptance criteria met: `filter[edible]=no` now returns the non-edible set, `filter[edible]=banana` returns 400 naming the field and value, suite green.

**CI**: green — RSpec pass (2m10s), RuboCop pass, Brakeman pass (run 33947959655).

**Local verification** (checked out branch in a worktree):
- `bundle exec rubocop` on the 3 changed files: no offenses.
- `bundle exec rspec spec/requests/api/v1/filter_validation_spec.rb`: 31 examples, 0 failures.
- `bundle exec rspec spec/requests/api`: 116 examples, 0 failures.
- `bundle exec rspec spec/requests` (full, incl. web): 190 examples, 0 failures — no regressions.

**Code**: the fix lives in the shared `Filterable` concern and uses generic `columns_hash` introspection rather than the `Species`-only `COLUMN_TYPES` constant, so it correctly generalizes to every boolean column reachable through `Filterable` (confirmed: `Species#edible`/`vegetable`/`flower_conspicuous`/`fruit_conspicuous`, `Plant#vegetable` via the shared `FILTERABLE_FIELDS` list) — matching the PR's claim of fixing the `Plant#vegetable` latent bug for free. Non-boolean columns are untouched (same `true`/`false`/`null` literal-string case as before). The new `InvalidFilterValueError` is rescued in `ApiController` and rendered with `render_error`, same envelope/style as the existing `UnknownQueryKeyError` path.

The `app/controllers/api/api_controller.rb` change isn't in the issue's `Touches:` line, but it's the necessary rescue wiring for the new error type raised in `filterable.rb` — tightly coupled to the stated scope, not scope creep.

Test spec is well-commented: it correctly notes `edible` isn't directly settable (derived from `vegetable`/`edible_part` via `complete_cache_fields`) and scopes assertions by `scientific_name` against the shared seeded dataset rather than asserting on raw counts.

Marking ready for merge train.

Issue #285, opened 2026-09-05
Name a licence for the data


lambda2:
Follow-up to community question #21 (2019, the oldest open issue), which is being closed with a pointer to https://trefle.io/terms.

Where things stand today:
- `LICENSE.md` is AGPLv3 and covers the **code**.
- `/terms` covers the **data** only by reference: comply with each upstream source's licence and credit Trefle. No named licence.
- `/terms` is linked from neither `/about` nor the documentation site.

Why it matters beyond legal tidiness: this is the blocker for the downloadable dataset (#81). Nobody republishes a dataset whose terms are "it depends on the source".

To decide:
- Whether Trefle's own aggregation can carry a named licence (ODbL and CC-BY-SA are the usual candidates for this kind of database), given the mix of upstream terms.
- The attribution text third parties should reproduce.
- Then: state it on /about, in the docs, and in the dump repository.

lambda2:
Decision material (2026-09-07). This issue is now the head of the `attribution-changes` chain — #318 #319 #320 #321 #322 implement whatever is decided here, and all but #318 are blocked on it.

## Upstream licences, verified today

| Source | Licence |
|---|---|
| WCVP / POWO (dataset releases) | **CC-BY 4.0** |
| GBIF Backbone | **CC-BY 4.0** |
| WFO Backbone | **CC0** |
| IPNI | CC-BY (to confirm with Kew) |
| USDA (residual) | Public domain |
| Wikipedia-derived content | CC-BY-SA ⚠ share-alike |
| PlantNet, CATMINAT/TRY residuals | unclear — see #284 |

The whole taxonomic backbone is CC-BY 4.0 or CC0: nothing upstream prevents naming a single open licence for the Trefle compilation, with per-record upstream credits (#318). The two watch-points are Wikipedia-derived text (share-alike contaminates) and the untraced residuals.

## The choice

- **CC-BY 4.0** — recommended. Version 4.0 explicitly covers the EU sui generis database right (the actual legal basis protecting the compilation, 15 years, directive 96/9/EC). Compatible with every upstream, universally understood, zero friction for integrators.
- **ODbL** — the OSM model: adds share-alike, so derived *databases* must stay open. More protective, more deterrent for commercial integrators, and adds compatibility questions. Worth it only if copyleft on the data is a goal in itself (the code is already AGPL-3.0).

[...]

PR #325, opened 2026-09-07
Terms acceptance at signup with a binding attribution clause

### LINKED ISSUE #320: Terms acceptance at signup with a binding attribution clause
The only *binding* attribution mechanism Trefle can have is contractual: the API requires a JWT token (`POST /api/auth/claim`), tokens require an account, and account creation is the moment where terms can be accepted. Copyright on facts is weak everywhere and the EU sui generis database right stops at the EU border; a terms-of-service clause accepted at signup binds every API consumer worldwide. Today nothing records that a user ever saw, let alone accepted, the terms.

## Scope

1. Migration: `terms_accepted_at` (datetime) and `terms_version` (string) on `users`.
2. Signup flow (Devise registration + OmniAuth GitHub callback): require an explicit terms-acceptance checkbox (unchecked by default), persist timestamp + version. Existing users: prompt for acceptance on next sign-in to the web app; do **not** cut API access retroactively.
3. Add the attribution clause to `/terms` (exact licence wording comes from #285):
   - applications displaying Trefle data must show "Data: Trefle.io (<licence>)" with a link;
   - publications should cite the canonical citation (the /citation page will carry formats).
4. Bump `terms_version`; the version string lives in config so future terms changes are a one-line bump.

## Acceptance criteria

- New signups (email and GitHub) cannot complete without accepting; `terms_accepted_at`/`terms_version` persisted.
- `/terms` contains the attribution clause and names the data licence.
- Existing users see a one-time acceptance prompt on the web; API tokens keep working regardless.
- Request/feature specs cover both signup paths; RuboCop/Brakeman green.

Touches: db/migrate, app/models/user.rb, Devise views/controllers (users/), app/views (terms page), config, spec
Blocked by: #285 (the clause names the licence).
Related: #286 (email confirmation decision touches the same signup flow — coordinate, don't merge).

Unblocked: #285 decided CC-BY 4.0 on 2026-09-07 — the clause wording can name it.

lambda2:
Closes #320

## Summary

- Migration adds `users.terms_accepted_at` / `users.terms_version`; `TERMS_VERSION` in `config/initializers/terms.rb` is the one-line bump point for future terms changes.
- Email sign-up requires an explicit, unchecked-by-default checkbox. Enforced via `User#enforce_terms_acceptance`, turned on only by the public registration form (`Users::RegistrationsController#build_resource`) — the admin back office (`management/users`) and GitHub OAuth account creation are unaffected by that validation.
- GitHub OAuth sign-up has no interactive step of its own before the account is created (`Users::OmniauthCallbacksController` creates the `User` directly in the callback). Rather than blocking creation, the account starts out pending (`terms_accepted_at` nil) and is picked up by the same gate used for pre-existing users.
- New `RequiresTermsAcceptance` concern, included in `ApplicationController` and `Explore::ExploreController`, redirects a signed-in user with missing/stale terms to `TermsAcceptancesController` before they can do anything else on the web app (browsing, submitting record corrections, profile, etc). Deliberately **not** included in `Management::ManagementController` (admin back office) or anywhere in the JSON API stack (`Api::ApiController`) — API tokens keep working regardless, including for the pre-existing `Api::V1::HomeController < ApplicationController` quirk (explicit guard).
- `/terms` now names the CC BY 4.0 data licence and the attribution clause ("Data: Trefle.io (CC BY 4.0)" + citation guidance), per #285 (closed, CC BY 4.0 decided 2026-09-07).

## Notes for the reviewer

- A subtle one caught in review: the shared checkbox partial originally used a bare `check_box_tag`, which (unlike `f.check_box`) has no hidden "0" fallback — an unticked box submitted no param at all, and the acceptance validator treats a genuinely unset attribute as not-applicable rather than "declined", silently letting signup through unchecked. Fixed by adding the hidden fallback field explicitly (see `app/views/shared/_terms_checkbox_field.html.erb`); covered by a feature spec.
- `db/schema.rb` in this PR reflects only this branch's migration — the shared local Postgres between concurrent worktree sessions can otherwise leak unrelated in-flight schema state into a fresh `db:migrate` dump, so I reset it to master + this migration only.

## Testing

- Full RSpec suite: 1059 examples, 0 failures.
- RuboCop: 0 offenses.
- Brakeman: 0 warnings.

Touches: db/migrate, app/models/user.rb, app/controllers (users/, terms_acceptances, concerns), app/views (devise/registrations, home/licence, terms_acceptances, shared), config, spec

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,11 +1,13 @@
 class ApplicationController < ActionController::Base
+  include RequiresTermsAcceptance
+
   protect_from_forgery with: :null_session
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :set_sentry_context
   before_action :set_meta
 
   def configure_permitted_parameters
-    added_attrs = %I[name account_type email password password_confirmation remember_me organization_name organization_url]
+    added_attrs = %I[name account_type email password password_confirmation remember_me organization_name organization_url accepts_terms]
     devise_parameter_sanitizer.permit :sign_up, keys: added_attrs
     devise_parameter_sanitizer.permit :account_update, keys: added_attrs
   end
--- a/app/controllers/concerns/requires_terms_acceptance.rb
+++ b/app/controllers/concerns/requires_terms_acceptance.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+# Shared by every controller that renders the public web app (see
+# ApplicationController and Explore::ExploreController): a signed-in visitor
+# whose terms_accepted_at/terms_version is missing or stale gets sent to
+# accept the current version (TermsAcceptancesController) before doing
+# anything else -- browsing explore/ pages, submitting a record correction,
+# managing their profile, etc.
+#
+# Deliberately NOT included in Management::ManagementController (the admin
+# back office -- gating it risks locking out an admin over a UX detail) or
+# anywhere in the JSON API (Api::ApiController is a separate
+# ActionController::API stack that never sees this concern; the
+# `controller_path.start_with?('api/')` guard below also covers the one
+# pre-existing quirk where Api::V1::HomeController inherits
+# ApplicationController directly instead of Api::ApiController). API access
+# is never gated on this (#320).
+module RequiresTermsAcceptance
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :require_terms_acceptance!
+  end
+
+  private
+
+  def require_terms_acceptance!
+    return unless user_signed_in?
+    return if devise_controller?
+    return if controller_path == 'terms_acceptances'
+    return if controller_path.start_with?('api/')
+    return if current_user.terms_up_to_date?
+
+    redirect_to new_terms_acceptance_path(return_to: request.fullpath)
+  end
+end
--- a/app/controllers/explore/explore_controller.rb
+++ b/app/controllers/explore/explore_controller.rb
@@ -1,5 +1,6 @@
 class Explore::ExploreController < ActionController::Base
   include Pagy::Backend
+  include RequiresTermsAcceptance
 
   layout 'application'
   before_action :generate_jwt
--- a/app/controllers/terms_acceptances_controller.rb
+++ b/app/controllers/terms_acceptances_controller.rb
@@ -1,0 +1,44 @@
+# frozen_string_literal: true
+
+# One-time prompt shown to a signed-in web visitor whose terms acceptance is
+# missing or stale (see RequiresTermsAcceptance and TERMS_VERSION in
+# config/initializers/terms.rb). Covers both pre-existing users (accepted an
+# older version, or never recorded one) and brand new GitHub OAuth sign-ups,
+# which have no interactive step to show a checkbox during the callback
+# itself. Never runs against the JSON API: Api::ApiController is a separate
+# ActionController::API stack that never includes RequiresTermsAcceptance, so
+# tokens keep working regardless of whether this prompt has been accepted
+# (#320).
+class TermsAcceptancesController < ApplicationController
+  before_action :authenticate_user!
+
+  def new
+    @return_to = safe_return_to(params[:return_to])
+  end
+
+  def create
+    if truthy_param?(params.dig(:user, :accepts_terms))
+      current_user.accepts_terms = true
+      current_user.save!
+      redirect_to safe_return_to(params[:return_to]), notice: 'Thanks -- your acceptance of the Terms of Use has been recorded.'
+    else
+      @return_to = safe_return_to(params[:return_to])
+      flash.now[:alert] = 'You must accept the Terms of Use to keep using the Trefle website.'
+      render :new, status: :unprocessable_entity
+    end
+  end
+
+  private
+
+  def truthy_param?(value)
+    ActiveModel::Type::Boolean.new.cast(value)
+  end
+
+  # `return_to` comes from an untrusted query/hidden-field param -- only
+  # honor same-site paths so it can't be turned into an open redirect.
+  def safe_return_to(path)
+    return root_path if path.blank? || !path.start_with?('/') || path.start_with?('//')
+
+    path
+  end
+end
--- a/app/controllers/terms_acceptances_controller.rb
+++ b/app/controllers/terms_acceptances_controller.rb
@@ -36,8 +36,14 @@ class TermsAcceptancesController < ApplicationController
 
   # `return_to` comes from an untrusted query/hidden-field param -- only
   # honor same-site paths so it can't be turned into an open redirect.
+  # Beyond a literal `//` prefix, reject backslashes and ASCII tab/newline/CR:
+  # browsers normalize `\` to `/` for special schemes, and strip tab/newline/CR
+  # entirely before resolving a Location header (WHATWG URL spec), so either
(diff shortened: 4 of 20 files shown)


lambda2:
🔎 claim-1788777829-2861

lambda2:
## Review: Fail

**CI**: green (`gh pr checks 325` — Brakeman, RSpec, RuboCop all pass).

**Contract (#320)** — most acceptance criteria are met: migration + `TERMS_VERSION` initializer, `enforce_terms_acceptance` scoped correctly to the public sign-up form only (verified `management/users_controller.rb`'s `user_params` doesn't permit `accepts_terms`/`enforce_terms_acceptance` — admin can't mass-assign either), GitHub OAuth starts pending and is caught by `RequiresTermsAcceptance` on next page load, the concern is correctly excluded from `Management::ManagementController` (`< ActionController::Base`, no include) and from the JSON API (`Api::ApiController < ActionController::API`, plus the explicit `controller_path.start_with?('api/')` guard for the `Api::V1::HomeController < ApplicationController` quirk — confirmed both by reading the code), `/terms` names CC BY 4.0 and the attribution clause, and both signup paths have request/feature spec coverage.

**One real bug — open redirect in `TermsAcceptancesController#safe_return_to`** (`app/controllers/terms_acceptances_controller.rb:38-42`):

```ruby
def safe_return_to(path)
  return root_path if path.blank? || !path.start_with?('/') || path.start_with?('//')
  path
end
```

The comment states the intent explicitly: "only honor same-site paths so it can't be turned into an open redirect." It doesn't. A `return_to` of `/\evil.com` (backslash, not a second forward slash) passes both checks — it starts with `/` and does not start with `//` — so `safe_return_to` returns it unchanged. Verified directly against this method (`rails runner`, no DB needed):

```
"/\\evil.com" => /\evil.com      # NOT caught
"//evil.com"  => root_path        # caught
```

For any "special" URL scheme (http/https — WHATWG URL spec), a backslash is normalized to a forward slash during parsing, including by browsers resolving a `Location:` header. `Location: /\evil.com` resolves to `//evil.com`, i.e. a protocol-relative redirect to `evil.com`. This is a well-known blocklist-bypass pattern (`start_with?('//')` alone is not a sufficient guard — `\/`, `//`, `/\`, and mixed variants all need to be rejected).

Exploit path: attacker sends a logged-in user (or one who signs in via the link, since the controller has `before_action :authenticate_user!` and Devise's stored-location will carry the query string back through) a link like `https://trefle.io/terms_acceptance/new?return_to=%2F%5Cevil.com`. After the victim accepts the terms (a totally reasonable action to take on your own site), `POST /terms_acceptance` redirects them straight to `evil.com` — a phishing vector, and notably the *same* one the code was explicitly written to prevent.

### What "fixed" looks like
- Reject any `return_to` that isn't a plain same-site path: at minimum also reject anything containing `\` (backslash), not just a literal `//` prefix. A robust version: parse with `URI.parse` and require `host` to be nil/blank, or use `Rack::Utils.escape_path` + reject on any `\` character before the `start_with?` checks.
- Add a regression spec asserting `return_to=/\evil.com` (and ideally `/\/evil.com`, `\/evil.com`) falls back to `root_path`, alongside the existing `//evil.com` case in `spec/requests/terms_acceptances_spec.rb`.

Please fix and re-request review — I'll leave this in draft with `needs-fixes` and won't fix it myself so the loop stays a loop.

lambda2:
🔧 claim-1788778905-20545

lambda2:
Fixed the open redirect: `safe_return_to` now also rejects any `return_to` containing a backslash or ASCII tab/newline/CR (`/\\evil.example`, `/\\/evil.example`, etc.), not just a literal `//` prefix -- these are the same browser-side normalization/stripping tricks that let a same-site-looking path resolve to a protocol-relative redirect. Added regression specs alongside the existing off-site guard test. RSpec (full `terms_acceptances_spec.rb`, 17/17) and RuboCop both pass on the changed files.

lambda2:
🔎 claim-1788779679-12639

lambda2:
## Review: Pass

**CI**: green on the fix commit `a96b42b` (`gh pr checks 325` — Brakeman, RSpec, RuboCop all pass).

**Fix verified.** The open redirect flagged in the prior review (`safe_return_to` letting `/\\evil.example` through because it starts with one `/` and not literally `//`) is fixed: the method now also rejects any `return_to` containing a backslash or ASCII tab/newline/CR, in addition to the existing blank/off-path/protocol-relative checks. I exercised `safe_return_to` directly (extracted the guard logic, ran it against `/\\evil.example`, `/\\/evil.example`, `//evil.example`, tab/newline variants, a percent-encoded `//`, and a Unicode division-slash lookalike) — every browser-normalizable bypass falls back to `root_path`; plain same-site paths and non-normalizable lookalikes (which the WHATWG URL spec doesn't rewrite to `/`) still pass through. Two regression specs were added for the backslash and mixed-slash cases.

**Contract (#320)** — re-verified independently in a fresh worktree at head commit `a96b42b`, not just re-reading the prior review:
- `terms_accepted_at`/`terms_version` migration + `TERMS_VERSION` initializer, minimal schema.rb diff (two columns only).
- `enforce_terms_acceptance` is set only in `Users::RegistrationsController#build_resource` (public sign-up); confirmed `management/users_controller.rb`'s `user_params` permits neither `accepts_terms` nor `enforce_terms_acceptance`, so the admin back office can't trip the validation via mass-assignment either.
- `RequiresTermsAcceptance` is included in `ApplicationController` and `Explore::ExploreController` only — confirmed absent from `Management::ManagementController` and never reachable from `Api::ApiController`'s stack; the concern's own `controller_path.start_with?('api/')` guard covers the `Api::V1::HomeController < ApplicationController` quirk.
- GitHub OAuth accounts start with `terms_accepted_at` nil and get caught by the same gate on next page load — no interactive step needed at callback time.
- `/terms` names CC BY 4.0 and the attribution clause ("Data: Trefle.io (CC BY 4.0)", citation guidance, binding-via-signup wording).
- Hidden-field fallback fix for the terms checkbox (bare `check_box_tag` has no unchecked fallback, unlike `f.check_box`) is real and covered by a feature spec.

**Verified myself** (fresh worktree, this machine's local Postgres/Redis/OpenSearch, `app/assets/builds` copied over since a bare worktree has no `yarn build`):
- `bundle exec rspec spec/requests/terms_acceptances_spec.rb spec/requests/users/omniauth_callbacks_spec.rb spec/requests/web/home_spec.rb spec/features/sign_in_spec.rb spec/models/user_spec.rb` → 48 examples, 0 failures.
- `bundle exec rubocop` on all touched files → 0 offenses.
- `bundle exec brakeman -q -w2` → 0 warnings.

**Scope**: diff matches the PR's `Touches:` line exactly (db/migrate + schema.rb, user.rb, controllers, views, config, spec). No drive-by changes.

No issues found.

lambda2:
> *This was generated by AI (watchdog).*


CI is red on the current head commit (`78b500a`): **Brakeman fails** (exit code 3) — 1 security warning, `EOLRails` check: "Support for Rails 8.0.5.1 ends on 2026-11-07" (`Gemfile.lock:412`). RSpec and RuboCop are green. Note the last human review (2026-09-07 11:19) passed this PR against an earlier commit (`a96b42b`) that was green on Brakeman — the head has since moved and now trips this warning. Labeling `needs-fixes` so the next worker session picks it up (likely: add an ignore entry for this EOL warning, matching the pattern already used in `config/brakeman.ignore`, or bump/pin as appropriate).

PR #329, opened 2026-09-07
Donate: rebuild the page to the institutional maquette

lambda2:
Implements `Trefle Donate.dc.html` from the design project (local reference: `.agents/design/home-2026/maquette-donate.html`), completing the public-page set (home #314/#315/#323, explore #327, about #317).

**Stacked on #323**, the tail of the #327 → #317 → #323 chain; bases retarget as they merge.

## What changes

- The donate page drops the old Bulma hero + sticky side menu for the maquette structure: serif "Support Trefle" header, then four hairline sections whose headings carry this page's own device — a **3px forest-green underline** (distinct from the home's hanging icons and the about page's target rail).
- **How to sponsor us**: three unnumbered step columns (reusing `.home-steps`), primary CTA "Sponsor on GitHub" with the GitHub brand icon, "Link my GitHub account" → profile.
- **Where the money goes**: monthly cost bars (reusing `.home-bar-*`; amounts €180/€60/€25/€10 come from the mockup — **please double-check them before this goes live**), and "Who covers it" with the **real sponsor count** from `users.sponsored_tier` (synced by the update_sponsors cron, cached 12h, number-aware verb). The mockup's "€165 covered from 41 sponsors" progress bar was placeholder arithmetic tier labels can't back, so it is deliberately not reproduced.
- **Company tiers**: Bronze/Silver/Gold cards (reusing `.about-tile`), invoice/bulk/dedicated mailto. The old page's **Sponsor Terms survive** as a `<details>` disclosure here — standing terms are not dropped silently, even though the mockup omits them.
- **Other ways to help**: four link tiles (Data → explore, Code → GitHub issues, Time → home volunteer section, Word → about licence/citation block).

## Verification

- Checked in Chrome against the maquette: header, underlined headings, steps, bars, tiers, terms disclosure, link tiles, footer.
- RSpec: 1053 examples, 0 failures (new request specs: sponsor count with verb agreement, tiers + terms presence). RuboCop: no offenses. `yarn build` clean.

Touches: app/controllers/home_controller.rb, app/views/home/donate.html.erb, app/assets/javascripts/, spec/requests/web/home_spec.rb

PR #327, opened 2026-09-07
Explore: rebuild the browse page to the institutional maquette

lambda2:
Recreates #316, which GitHub auto-closed (not merged) when #315's base branch was deleted at merge and the retarget failed. Same branch, same commits, base now master. See #316 for the original description and review context.

Implements `Trefle Explore.dc.html` (local reference: `.agents/design/home-2026/maquette-explore.html`): full-bleed explore browse page with filter chips (rank / top families / with-images / edible / vegetable), sort select (featured, name, most complete, recently corrected), completion meta row on cards, empty state, contribute band. Filters translate to searchkick where clauses in search mode; three new indexes back the sorts and the family filter (migration included).

Verified in Chrome on live data; RSpec 1049/0 at commit time (10 new request specs), RuboCop clean.

Touches: app/controllers/explore/, app/helpers/, app/views/explore/species/, app/assets/javascripts/sections/_explore.scss, db/migrate/, spec/requests/web/

Chain: this PR ← #317 (About) ← #323 (home design pass).

PR #323, opened 2026-09-07
Home design pass: aligned headings, no step numbers, wrapped commands, AA muted grey

lambda2:
Ergonomics and de-templating pass on the home page, checked visually in Chrome at each step.

**Stacked on #317** (about), the tail of the #315 → #316 → #317 chain; bases retarget as they merge.

## Changes

- **Section title alignment**: `.home-section-heading` shifts left by exactly the icon + gap width (2.5rem) on viewports ≥ 1080px, so the serif H2 text lines up with the section body's left axis and the duotone icon hangs in the margin. Narrower screens keep the inline icon (no room in the margin).
- **For developers**: the `01 / 02 / 03` step numbers are removed. The step headings are the labels; the `<ol>` keeps the sequence for assistive tech. (`home-steps__index` CSS removed with it.)
- **No more stray horizontal scrollbars**: the curl sample wraps (`pre-wrap` + `overflow-wrap: anywhere`), and the live-demo command line uses `wrapLongLines` — react-syntax-highlighter pins `white-space: pre` inline on its `<code>` tag, so that prop is the only clean way (verified in the DOM: 3 wrapped lines, zero overflow). The JSON response block deliberately keeps its scroll: payloads are data, not prose.
- **Accessibility**: muted secondary text (`#7a7a7a`) measures **4.3:1 on white, under WCAG AA's 4.5:1**. It moves to a single `$text-muted` variable at `#6e6e6e` (5.1:1), swapped across all section stylesheets (home, explore, about, navbar, footer) so the grey stays uniform sitewide; legacy `#6f6f6f` literals folded into the same variable.

## Verification

- Chrome pass at desktop width: headings aligned on the content axis on all six sections (including the two-column Support section), numbering gone, terminal and curl blocks scrollbar-free.
- RSpec: 1051 examples, 0 failures. RuboCop: no offenses. `yarn build` clean.

Touches: app/views/home/_developers.html.erb, app/assets/javascripts/sections/, app/assets/javascripts/home/codesandbox.js

PR #317, opened 2026-09-07
About: rebuild the page to the institutional maquette

lambda2:
Implements `Trefle About.dc.html` from the design project (local reference: `.agents/design/home-2026/maquette-about.html`), continuing the identity work from #314/#315/#316.

**Stacked on #316** (explore), itself stacked on #315 (the `full_bleed` layout opt-out); bases retarget automatically as they merge.

## What changes

- The about page drops the old Bulma hero + sticky side menu for the maquette structure: serif "About Trefle" heading + factual lead (Mashum link), an anchor nav between hairlines, then six stacked sections separated by hairlines.
- **Sections**: Data sources (7-logo grayscale grid with real alt texts), The team, Pricing and limits (three bordered tiles: 60 req/min free, 600 for sponsors, bulk export for institutions), Licence and citation (a ready-to-copy `<pre>` citation dated per request — the page has no fragment cache), the (very short) privacy policy, Contact (e-mail / Discord / GitHub tiles with brand-colored icons).
- **The team tiles carry computed numbers** instead of copy that would drift: admin count, distinct correction reviewers (`accepted_by`), distinct contributors this year — cached for a day.
- `section:target` shows the maquette's 2px forest-green left rail; anchor navigation scrolls smoothly behind a `prefers-reduced-motion` guard.
- Dropped: the "Beta stage" section (that honesty now lives in the home Data section) and the Twitter contact tile.
- The home stat tiles gain a `--compact` modifier reused here.

## Verification

- Checked in Chrome against the maquette: header + anchor nav, all six sections, the `:target` rail on `/about#pricing`, brand icons, footer.
- RSpec: 1051 examples, 0 failures — including new request specs for the team numbers, rate-limit tiles, per-request citation date, and an anchor-nav integrity check (every nav href resolves to a section id). Feature spec updated to the new copy. RuboCop: no offenses. `yarn build` clean.

Touches: app/controllers/home_controller.rb, app/views/home/about.html.erb, app/assets/javascripts/, spec/

PR #330, opened 2026-09-07
Chrome consistency: aligned navbar, square cards, institutional account page

lambda2:
Three sitewide consistency fixes requested after reviewing the pages side by side. **Stacked on #329**, tail of the #327 → #317 → #323 → #329 chain.

## Changes

- **Navbar alignment**: the header hairline spans the full viewport, and the nav content sits in the same 1040px + 1.5rem-padding container as every page body — the brand logo now shares the pages' left axis (verified against the explore search field in Chrome). Root cause of the old misalignment: Bulma's responsive `.container` caps use `:not()` selectors that outweigh a plain class override, so the previous `max-width: 1040px` never applied; the header no longer uses Bulma's `.container` at all. The brand link drops Bulma's leading item padding.
- **Square cards everywhere**: the species card photo container takes `aspect-ratio: 1/1` with the existing `object-fit: cover` image — portrait herbarium scans and landscape field photos crop to the same square, and the leaf placeholder fills the identical frame instead of collapsing into a thin strip (the genus-page bug in the screenshots). Applies to the explore grid and the genus pages, which share the partial.
- **Account page** joins the institutional structure (same header + hairline sections as the about page, reusing its classes): serif "Your account" with the signed-in e-mail in mono, anchor nav, Access token section (mono readonly field, click-to-select, rate-limit line — the sponsor link now points at the donate page's sponsor section instead of the about anchor removed in #317), a label/value details grid (rows shown only when filled), and a separate "Sign out & deletion" section that isolates the destructive action and states what deletion does. Page title added.

## Verification

- Chrome pass: navbar/content left axis aligned on explore + profile, full-width hairline, square cropping on mixed portrait/landscape/placeholder rows (real Fabaceae hybrids data), account page top to bottom.
- RSpec: 1053 examples, 0 failures (account feature spec updated to the new section names, token-input assertion still passing). RuboCop: no offenses. `yarn build` clean.

Touches: app/views/layouts/_navbar.html.erb, app/views/profile/, app/controllers/profile_controller.rb, app/assets/javascripts/sections/, spec/features/account_spec.rb

### DIFF
--- a/app/assets/javascripts/application.scss
+++ b/app/assets/javascripts/application.scss
@@ -14,6 +14,7 @@
 @import 'sections/explore';
 @import 'sections/about';
 @import 'sections/donate';
+@import 'sections/profile';
 @import 'sections/navbar';
 @import 'sections/footer';
 
--- a/app/assets/javascripts/sections/_explore.scss
+++ b/app/assets/javascripts/sections/_explore.scss
@@ -170,14 +170,16 @@
   align-items: stretch;
   flex-direction: column;
 
+  // Every card shows the same square frame: portrait herbarium scans and
+  // landscape field photos are cropped by the object-fit:cover img below,
+  // and the leaf placeholder fills the identical square.
   aside {
-    min-width: 100px;
-    min-height: 150px;
-    height: 100%;
+    aspect-ratio: 1 / 1;
     display: flex;
     margin: 10px 0;
     border-radius: 3px;
     overflow: hidden;
+    width: 100%;
   }
 
   main {
--- a/app/assets/javascripts/sections/_navbar.scss
+++ b/app/assets/javascripts/sections/_navbar.scss
@@ -6,17 +6,30 @@
 // stays Bulma's own navbar/navbar-burger/navbar-menu markup so the existing
 // mobile burger toggle (application.js) keeps working unmodified — only the
 // look is restyled here.
-// Bulma's own .container caps at 960px on desktop (see bulma/elements/
-// container.sass's $desktop - $container-offset); the maquette wants 1040px
-// (see #304 / #312 review) so this overrides it specifically for the navbar,
-// without touching .container's default elsewhere on the site.
-.site-navbar-container {
-  max-width: $layout-max-width;
+// The hairline spans the full viewport; the nav content sits in the same
+// 1040px + 1.5rem-padding container as every page body, so the brand and the
+// first heading below share a left axis. No Bulma .container here: its
+// responsive caps use :not() selectors that outweigh a plain class override.
+.site-header {
+  background: #fff;
+  border-bottom: 1px solid #dbdbdb;
+
+  &__inner {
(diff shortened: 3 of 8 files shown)


lambda2:
🔎 claim-1788855683-24817

lambda2:
## Review: Pass

**CI**: green on head commit `d9bbf7c` (`gh pr checks 330` — Brakeman, RSpec, RuboCop all pass; `headRefOid` matches the checked-out commit).

**Contract** — no linked issue; this is the tail of the #327 → #317 → #323 → #329 institutional-maquette chain, so the bar is internal consistency with the sibling rebuilds plus the three fixes described in the PR body. Verified each:

- **Navbar**: `.site-header`/`.site-header__inner` replaces `.site-navbar-container`; confirmed the new container shares `$layout-max-width` (1040px) + `1.5rem` padding with `.about__header`/`.about__body` and the other page bodies, and `.brand-link` zeroes Bulma's `0.75rem` navbar-item padding so the logo lands on the same left edge. Matches the stated root cause (Bulma's `.container` `:not()` caps beating a class override).
- **Square cards**: `.species-grid-item aside` now uses `aspect-ratio: 1/1` + `width: 100%` instead of `min-width/min-height/height:100%`. Confirmed via `_species_item` partial that this is shared by both `explore/species/index` and `explore/genus/index`+`show`, so the genus-page placeholder-collapse bug is actually fixed, not just papered over on one page.
- **Account page**: rebuilt on `.about`/`.about-section` classes (all present in `_about.scss` via `&__x` nesting — checked, since a flat grep initially missed them). `donate_path(anchor: 'sponsor')` correctly targets `#sponsor` in the #329 donate page (old target was the `about#support` anchor removed in #317). `@page_title = 'Your account'` follows the same pattern as the other page controllers. `provide :full_bleed, '1'` matches the convention used by home/about/donate/species-index.

**Local verification** (not just trusting CI):
- `bundle exec rspec spec/features/account_spec.rb spec/features/ spec/requests/web/explore_spec.rb` → 45 examples, 0 failures.
- `bundle exec rubocop app/controllers/profile_controller.rb spec/features/account_spec.rb` → no offenses.
- `yarn build` → compiles clean (pre-existing bundle-size warnings only, unrelated to this diff).

**Non-blocking note**: `profile-rate--sponsor` is added as a class on the sponsor-tier `<p>` in `profile/index.html.erb` but has no corresponding rule in `_profile.scss` (only `.profile-rate` and `.profile-rate__crown` are defined) — harmless (falls back to the base `.profile-rate` styling) but looks like a leftover from an intended sponsor-specific style. Also, the new "Sign out" `button_to` drops the `data: { confirm: ... }` the old link had — sign-out isn't destructive so this doesn't block, just flagging the behavior change since "Sign out & deletion" pairs it visually with the account-deletion button, which does keep its confirm.

Both are cosmetic/UX polish, not contract breaks — not blocking.

PR #333, opened 2026-09-08
Give each worktree its own test database

lambda2:
Parallel sessions run the suite from several git worktrees against one local Postgres. `db:test:prepare` drops and recreates the test database, so two runs finishing at the same moment race on it and one dies:

```
PG::UniqueViolation: duplicate key value violates unique constraint "pg_database_datname_index"
DETAIL:  Key (datname)=(trefle_test) already exists.
Tasks: TOP => db:test:load_schema => db:test:purge
```

The drop and the create are not atomic together — the second run recreates the database in between the first one's `DROP` and its `CREATE`. Observed today; the same shared-Postgres hazard already showed up once as schema drift leaking between concurrent worktrees.

## Change

- `config/database.yml` reads the test database name from `PG_TEST_DATABASE`.
- `.agents/verify.sh` derives one per worktree from the directory name (lowercased, non-alphanumerics collapsed to `_`, truncated to fit Postgres' 63-byte identifier limit).

Unset — CI, and a plain run in the main checkout — keeps `trefle_test`, so nothing about the existing setup changes. CI passes no such variable, and the main checkout is detected by comparing canonical paths against the git common dir.

`database.yml` uses plain Ruby (`ENV[...].to_s.empty? ? ... : ...`) rather than `.presence`, because the file is parsed early enough that ActiveSupport's core extensions are not guaranteed to be loaded.

## Verification

- Full `verify.sh` in a worktree: RuboCop clean, `1092 examples, 0 failures`, on its own database `trefle_test_fix_test_db_per_worktree`.
- Confirmed the main checkout takes the other branch and stays on `trefle_test`.
- Re-ran after fixing a trailing `_` in the derived name (`basename`'s newline was being folded into the slug by `tr`).

## Known limitation

Per-worktree databases are created but never dropped, so they accumulate as worktrees come and go. They are cheap and empty-ish, but a periodic `DROP DATABASE` sweep over `trefle_test_%` would be worth adding if the count becomes a nuisance. Not done here to keep this change to the race it fixes.

Touches: `config/database.yml`, `.agents/verify.sh`

### DIFF
--- a/.agents/verify.sh
+++ b/.agents/verify.sh
@@ -6,12 +6,25 @@
 set -euo pipefail
 cd "$(dirname "$0")/.."
 
-main_checkout=$(dirname "$(git rev-parse --path-format=absolute --git-common-dir)")
+main_checkout=$(cd "$(dirname "$(git rev-parse --path-format=absolute --git-common-dir)")" && pwd -P)
 for f in .env.local .env.test.local .env.development.local; do
   if [ ! -f "$f" ] && [ -f "$main_checkout/$f" ]; then
     cp "$main_checkout/$f" "$f"
   fi
 done
+
+# Several sessions run this script at once, from different worktrees, against one
+# local Postgres. `db:test:prepare` drops and recreates the test database, so two
+# runs finishing together race on it and one dies with
+#   PG::UniqueViolation ... duplicate key value violates unique constraint
+#   "pg_database_datname_index" ... Key (datname)=(trefle_test) already exists.
+# Give each worktree its own database (see config/database.yml). The main checkout
+# and CI set nothing and keep plain `trefle_test`, so nothing else changes.
+if [ "$(pwd -P)" != "$main_checkout" ]; then
+  slug=$(printf '%s' "$(basename "$(pwd -P)")" | tr '[:upper:]' '[:lower:]' | tr -c 'a-z0-9' '_' | cut -c1-45)
+  export PG_TEST_DATABASE="trefle_test_${slug}"
+  echo "verify: using test database ${PG_TEST_DATABASE}"
+fi
 
 bundle check >/dev/null 2>&1 || bundle install --quiet
 
--- a/config/database.yml
+++ b/config/database.yml
@@ -59,7 +59,11 @@ development:
 # Do not set this db to the same as development or production.
 test:
   <<: *default
-  database: trefle_test
+  # Agent sessions run the suite from several git worktrees against one local
+  # Postgres, and `db:test:prepare` drops and recreates this database — two runs
+  # finishing at once collide (PG::UniqueViolation on pg_database). PG_TEST_DATABASE
+  # lets each checkout claim its own. Unset (CI, a plain local run) keeps trefle_test.
+  database: <%= ENV['PG_TEST_DATABASE'].to_s.empty? ? 'trefle_test' : ENV['PG_TEST_DATABASE'] %>
 
 # As with config/secrets.yml, you never want to store sensitive information,
 # like your database password, in your source code. If your source code is


lambda2:
🔎 claim-1788862088-18941

lambda2:
## Review: Pass

**CI**: green on head commit `cd1a721` (`gh pr checks 333` — Brakeman, RSpec, RuboCop all pass).

**Contract** — no linked issue; this is a self-scoped infra fix, contract is the PR body's own description of the race and its fix. Verified each claim:

- `config/database.yml`: `database: <%= ENV['PG_TEST_DATABASE'].to_s.empty? ? 'trefle_test' : ENV['PG_TEST_DATABASE'] %>` — plain Ruby ternary rather than `.presence`, correct given `database.yml` is parsed before ActiveSupport core extensions load.
- `.agents/verify.sh`: `main_checkout` computation now goes through `cd ... && pwd -P` (previously just `dirname` on the git-common-dir path, unresolved) — necessary because the new `if [ "$(pwd -P)" != "$main_checkout" ]` guard needs both sides canonicalized or the main checkout would misdetect itself as a worktree on any symlinked path.
- Slug derivation (`tr -c 'a-z0-9' '_' | cut -c1-45`) uses `printf '%s'` instead of `echo`, avoiding the trailing-newline-becomes-`_` bug the PR mentions fixing. `trefle_test_` (12) + 45-char slug = 57 bytes, under Postgres' 63-byte identifier limit.
- Grepped the repo for other hardcoded `trefle_test` references (`scripts/`, `.rb`, `.rake`, `.github/workflows/ci.yml`) — none exist outside the two touched files, so nothing else silently keeps assuming the old fixed name.
- `.github/workflows/ci.yml` sets no `PG_TEST_DATABASE` — confirmed CI keeps the `trefle_test` branch.

**Local verification** (worktree at the PR's own branch, `.claude/worktrees/fix-test-db-per-worktree`, already on head commit):
- `PG_TEST_DATABASE=trefle_test_fix_test_db_per_worktree RAILS_ENV=test bin/rails db:test:prepare` → succeeds, and `psql -l` confirms both `trefle_test` and `trefle_test_fix_test_db_per_worktree` coexist as separate databases.
- `bundle exec rspec spec/models/species_spec.rb` against that database → 12 examples, 0 failures.
- Main checkout (different branch, pre-fix `database.yml`) confirmed still resolves to plain `trefle_test` with `PG_TEST_DATABASE` unset — consistent with the "nothing changes for a plain local run" claim.

**Known limitation** (author already flags this in the PR body, not a contract break): per-worktree databases are created but never dropped, so they accumulate as worktrees churn — a periodic `DROP DATABASE trefle_test_%` sweep would be worth adding later, out of scope for this race fix.

lambda2:
Second occurrence today, this one caught with the evidence in hand.

A session on another branch failed `verify.sh` with two rswag failures in `spec/requests/api/v1/15_family_spec.rb`. Not a code regression — the shared test database had been left half-built by the same race:

```
db:test:purge → PG::UniqueViolation ... Key (datname)=(trefle_test) already exists
...
You have 69 pending migrations
```

`db:drop` then refused outright:

```
PG::ObjectInUse: database "trefle_test" is being accessed by other users
DETAIL:  There is 1 other session using the database.
```

— a concurrent `rspec` was running against `trefle_test` at that moment. Waiting for it to finish, rebuilding the database (`db:drop db:create db:schema:load`) and re-running gave `1064 examples, 0 failures` with no code change at all. So the failure mode is not just a noisy abort: it can corrupt the database into a state that produces *plausible-looking spec failures* on unrelated specs, which is considerably worse to debug.

Worth recording for review: this change does cover the fleet, not just manually created worktrees. `stop-verify.sh` resolves its root with `git rev-parse --show-toplevel` and runs `cd "$root" && .agents/verify.sh`, so a session that has entered a worktree runs verify from that worktree — `pwd -P` differs from the git common dir, and the suffix applies. Only the main checkout keeps plain `trefle_test`, and that one belongs to a single interactive session.

Residual case this does not cover: two sessions running concurrently in the main checkout itself. Rare by construction, and closing it would mean a lock rather than a per-directory name.

PR #336, opened 2026-09-08
Fix filter_not[image_url] returning nothing on the search endpoints

### LINKED ISSUE #280: filter_not[image_url]=null returns nothing on the search endpoints
Implements the fix for community report #73 (2020), reproduced against production on 2026-09-04: `/api/v1/plants/search?q=coco` returns 247 results, and the same query with `filter_not[image_url]=null` returns **0**.

Cause: the AR-backed list endpoints have an explicit scope translating `image_url` to the real column `main_image_url` (`app/models/concerns/scopes/species.rb`). The Searchkick path has no equivalent: `search_params` in `app/controllers/api/api_controller.rb` passes `image_url` straight into Searchkick's `where:`, and the index has no such field (`search_data` only carries `main_image_url`).

Scope:
- Make the search path agree with the SQL path. Either index an `image_url` field, or translate the key before it reaches Searchkick. Prefer whichever keeps one source of truth for the mapping rather than two lists that can drift again.
- Audit the other filterable keys for the same mismatch between the SQL path and the search path, and report what you find in the PR.
- Request specs on the search endpoints for the filter, plus one for whatever other mismatch the audit turns up.

Closes #73.
Touches: app/controllers/api/api_controller.rb, app/models/concerns/search/, spec/requests
Blocked by: #216 (same API controller zone)

lambda2:
Closes #280.

## Bug

`/api/v1/plants/search?q=coco` returns real results; the same query with `filter_not[image_url]=null` returned **0**. The AR-backed list endpoints (`GET /api/v1/species`, `GET /api/v1/plants`) already got this right via a scope in `app/models/concerns/scopes/species.rb` that translates the public key `image_url` to the real column `main_image_url`. The Searchkick-backed `/search` endpoints had no equivalent: `search_params` (`app/controllers/api/api_controller.rb`) passed `image_url` straight into Searchkick's `where:`, and the index has no such field (only `main_image_url`), so the filter silently matched nothing.

## Fix

Added one mapping, `Scopes::Species::FIELD_ALIASES`, used by both paths instead of two lists that can drift apart again:
- the AR scope (`filter_not_by_image_url`) now reads the column name from it instead of a hardcoded literal
- `Api::ApiController#search_params` translates `filter`/`filter_not`/`range`/`order` keys through it before building the Searchkick `where:`/`order:` hash, so both `/api/v1/species/search` and `/api/v1/plants/search` (which share `search_params`) are fixed together

Landed in `scopes/species.rb` + `api_controller.rb` rather than `app/models/concerns/search/` (the issue's guess at scope): translating the key before it reaches Searchkick, per the issue's own preferred option, didn't need any change to what gets indexed.

## Audit: other filterable keys with the same mismatch

Went through `FILTERABLE_FIELDS`, `FILTERABLE_NOT_FIELDS`, `ORDERABLE_FIELDS`, `RANGEABLE_FIELDS` against what's actually indexed (`Search::Species#search_data` = `attributes` merged with a few explicit overrides for bitmask/array fields). Everything else is a real `species` column (or a denormalized one, e.g. `genus_name`/`family_name`) that's present in `attributes` under the same name, so it lines up on both paths.

One more mismatch of the exact same shape: **`establishment`** (`filter[establishment]`, in `FILTERABLE_FIELDS`). On the AR path it's a dedicated scope joining `species_distributions`. It's never indexed by Searchkick at all, so `filter[establishment]=native` on `/search` silently returns zero results too, same as `image_url` did.

**Not fixed here** — unlike `image_url` there's no single real column to alias to; a real fix means indexing `species_distributions` data into Searchkick, which is a data-pipeline change out of this issue's scope. Pinned with a characterization spec (`spec/requests/api/v1/search_filter_field_aliases_spec.rb`) documenting the current (silently-wrong) behavior so it's easy to find and doesn't regress further unnoticed. Flagging it here as the follow-up worth filing.

## Tests

- Two request specs reproduce the original bug against real Searchkick/OpenSearch (`search: true`, forcing synchronous indexing since the model's `callbacks: :queue` mode doesn't index inline during specs) and confirm the fix on both `/api/v1/species/search` and `/api/v1/plants/search`.
- One characterization spec pins the known `establishment` gap.
- Verified all three regress to failing against the pre-fix code (temporarily reverted locally) before confirming green with the fix.
- Full suite: 1095 examples, 0 failures. RuboCop: 0 offenses.

Touches: app/controllers/api/, app/models/concerns/scopes/, spec/requests

### DIFF
--- a/app/controllers/api/api_controller.rb
+++ b/app/controllers/api/api_controller.rb
@@ -230,29 +230,39 @@ module Api
 
     # @TODO ugly one
     # Turn query params into elasticsearch query
-    def search_params(filter_not_fields: [], filter_fields: [], order_fields: [], range_fields: [])
+    #
+    # field_aliases translates a public filter/order/range key into the field
+    # name actually indexed by Searchkick, for the keys where they differ
+    # (e.g. `image_url` -> `main_image_url`, see Scopes::Species::FIELD_ALIASES)
+    # -- the AR-backed path (Filterable#filter_with et al.) does the same
+    # translation through its own scope, so the two paths stop drifting (#280).
+    def search_params(filter_not_fields: [], filter_fields: [], order_fields: [], range_fields: [], field_aliases: {})
       where = {}
       order = nil
       if params[:filter_not].is_a?(ActionController::Parameters)
         params[:filter_not].permit(filter_not_fields).slice(*filter_not_fields).each do |field, value|
-          where[field] ||= {}
-          where[field][:not] = value&.split(',')&.map {|e| e.blank? || e == 'null' ? nil : e }
+          indexed_field = field_aliases.fetch(field, field)
+          where[indexed_field] ||= {}
+          where[indexed_field][:not] = value&.split(',')&.map {|e| e.blank? || e == 'null' ? nil : e }
         end
       end
       if params[:filter].is_a?(ActionController::Parameters)
         params[:filter].permit(filter_fields).slice(*filter_fields).each do |field, value|
-          where[field] = value.split(',')
+          where[field_aliases.fetch(field, field)] = value.split(',')
         end
       end
       if params[:range].is_a?(ActionController::Parameters)
         params[:range].permit(range_fields).slice(*range_fields).each do |field, value|
           min, max = value.split(',')
-          where[field] ||= {}
-          where[field][:gte] = min if min.present?
-          where[field][:lte] = max if max.present?
+          indexed_field = field_aliases.fetch(field, field)
+          where[indexed_field] ||= {}
+          where[indexed_field][:gte] = min if min.present?
+          where[indexed_field][:lte] = max if max.present?
         end
       end
-      order = params[:order].permit(order_fields).slice(*order_fields).to_unsafe_hash if params[:order].is_a?(ActionController::Parameters)
+      if params[:order].is_a?(ActionController::Parameters)
+        order = params[:order].permit(order_fields).slice(*order_fields).to_unsafe_hash.transform_keys {|field| field_aliases.fetch(field, field) }
+      end
       {
         where: where,
         order: order
--- a/app/controllers/api/v1/plants_controller.rb
+++ b/app/controllers/api/v1/plants_controller.rb
@@ -92,7 +92,8 @@ class Api::V1::PlantsController < Api::ApiController
       filter_not_fields: Api::V1::SpeciesController::FILTERABLE_NOT_FIELDS,
       filter_fields: Api::V1::SpeciesController::FILTERABLE_FIELDS,
       order_fields: Api::V1::SpeciesController::ORDERABLE_FIELDS,
-      range_fields: Api::V1::SpeciesController::RANGEABLE_FIELDS
+      range_fields: Api::V1::SpeciesController::RANGEABLE_FIELDS,
+      field_aliases: ::Scopes::Species::FIELD_ALIASES
     )
     options = {
       where: { main_species_id: nil }.merge(search_params[:where]),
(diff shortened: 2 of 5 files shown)


lambda2:
🔎 claim-1788865538-17745

lambda2:
## Review: Pass

**CI**: green on head commit `ae685d6` (`gh pr checks 336` — Brakeman, RSpec, RuboCop all pass).

**Contract** (#280, unblocked — #216 closed):
- Root cause matches the issue exactly: the AR-backed list endpoints already alias public key `image_url` -> column `main_image_url` via a scope in `scopes/species.rb`; the Searchkick-backed `/search` endpoints passed `image_url` straight into `where:` with no equivalent indexed field (verified in `search/species.rb#search_data`: it's `attributes` + a fixed set of overrides, no `image_url`, only `main_image_url` as a plain column).
- Fix is a single source of truth (`Scopes::Species::FIELD_ALIASES`) consumed by both paths — the AR scope (`filter_not_by_image_url`) reads the column from it instead of a literal, and `search_params` translates `filter`/`filter_not`/`range`/`order` keys through it before building Searchkick's `where:`/`order:`. Matches the issue's explicit preference ("keeps one source of truth... rather than two lists that can drift again").
- Confirmed both `/api/v1/species/search` and `/api/v1/plants/search` are fixed together: both share `search_params`, and both call sites (`species_controller.rb`, `plants_controller.rb`) now pass `field_aliases: ::Scopes::Species::FIELD_ALIASES` — grepped for other `search_params` callers, there are none, so nothing was missed.
- `FILTERABLE_FIELDS`/`FILTERABLE_NOT_FIELDS`/`ORDERABLE_FIELDS`/`RANGEABLE_FIELDS` are all `%w[]` string arrays, matching `FIELD_ALIASES`'s string keys — the `.fetch(field, field)` lookups line up on type, including the `order` case which goes through `transform_keys` on the permitted hash (also string-keyed).
- Audit claim spot-checked: `establishment` is filtered via `joins(:species_distributions).where(species_distributions: { establishment: v })` — confirmed it has no equivalent anywhere in `search_data`/`additional_search_data`, so it has the exact same silent-failure shape as `image_url` did, correctly identified as "not fixed here" (no single column to alias to) and pinned with a characterization spec rather than silently left unnoticed.
- `Touches:` line matches the diff exactly (`app/controllers/api/`, `app/models/concerns/scopes/`, `spec/requests`) — nothing outside stated scope.

**Local verification** (existing worktree at the PR's branch, already on head commit `ae685d6`):
- `bundle exec brakeman -q` → 0 warnings.
- Targeted specs (`search_filter_field_aliases_spec.rb`, `3_species_spec.rb`, `2_plants_spec.rb`, `species_controller_spec.rb`) → 228 examples, 0 failures.
- Full suite → 1095 examples, matching the PR's claimed count exactly; 2 pre-existing failures in `spec/workers/api_sweep_worker_spec.rb` (cursor position assertions) are unrelated to this diff (that worker/spec isn't touched by the PR at all) and non-deterministic — reran the file alone and got a *different* failure signature (`nil` vs `5` instead of `5` vs `10`), confirming pre-existing flakiness rather than a regression from this change.
- Reproduced the bug directly rather than trusting the PR's word for it: reverted the 4 changed source files (kept the new spec) and reran `search_filter_field_aliases_spec.rb` — the two `image_url` specs fail exactly as described (`expected [] to include "Zzztestus withphotoae"`), confirming they exercise real Searchkick/OpenSearch and genuinely catch the regression. Restored the fix afterward; clean working tree.

No gaps found against the acceptance criteria. Converting to ready for merge-train.

PR #338, opened 2026-09-08
Management: fix unnamespaced route helpers and friendly_id lookups on show pages

### LINKED ISSUE #332: Management: /management/families/:id 500s on an unnamespaced families_path
`/management/families/:id` raises on render, so the family page of the back-office is unreachable.

## Symptom

Sentry [API-86](https://trefle.sentry.io/issues/API-86) — environment `development`, release `c078ced` (current `master` HEAD), hit at `http://127.0.0.1:3232/management/families/50`.

```
ActionView::Template::Error: undefined local variable or method 'families_path'
  app/views/management/families/show.html.erb:34
```

## Root cause

`app/views/management/families/show.html.erb:34`:

```erb
<%= link_to 'Back', families_path %>
```

There is no top-level `families_path`. `resources :families` is declared twice, both times inside a namespace:

- `config/routes.rb:28` — under `namespace 'api' > namespace 'v1'`, giving `api_v1_families_path`
- `config/routes.rb:118` — under `namespace 'management'`, giving `management_families_path`

The preceding line of the same view already uses the namespaced form (`edit_management_family_path`), which is why line 33 renders and line 34 blows up.

Not caught by production yet only because the back-office is admin-only and rarely exercised — it is on `master`, so it will ship in the next deploy.

## Proposed fix

`families_path` → `management_families_path`.

There is no spec coverage for `management/families` at all (no `spec/requests/management/`, no reference to `management_family` anywhere under `spec/`), which is why CI is green on a page that cannot render. Worth adding a request spec for index and show while fixing, rather than the one-line change alone.

## Acceptance criteria

- `GET /management/families/:id` renders 200 for an admin.
- A request spec covers `management/families#index` and `#show`.
- The other `management/*` show views are checked for the same unnamespaced-helper mistake.

lambda2:
Closes #332

## What was broken

`app/views/management/families/show.html.erb` called `families_path`, which doesn't exist at the top level — `resources :families` is only declared inside the `api/v1` and `management` namespaces (giving `api_v1_families_path` / `management_families_path`). Visiting `/management/families/:id` raised `NoMethodError` on render (Sentry API-86).

## Fix

- `families_path` → `management_families_path` in the families show view.
- Audited every other `management/*` show view for the same scaffold mistake (per the issue's acceptance criteria) and found the identical bug in 11 more: `division_classes`, `division_orders`, `divisions`, `foreign_sources`, `foreign_sources_plants`, `kingdoms`, `major_groups`, `sessions`, `species_images`, `species_proposals`, `subkingdoms`. Fixed all of them the same way. `genuses`, `plants`, `species` and `users` show views already used correctly namespaced helpers.

## Related bug found while fixing

Fixing the families route helper wasn't enough on its own to make the page reachable via normal navigation: `Family` (and 6 other resources — `divisions`, `division_classes`, `division_orders`, `foreign_sources`, `kingdoms`, `subkingdoms`) use FriendlyId slugged, but their controllers looked records up with `Model.find(params[:id])`. Every view link (`link_to 'Show', [:management, family]`, `edit_management_family_path(family)`, ...) passes the slug via `to_param`, not the numeric id, so `Model.find(<slug>)` raised `ActiveRecord::RecordNotFound` — the show/edit pages stayed unreachable by clicking through the UI even after the route helper fix, only "working" if you hand-typed the numeric id in the URL (as the original Sentry report did). Switched those 7 controllers to `Model.friendly.find(params[:id])`, matching the pattern already used by the (untouched) `genuses`/`plants`/`species` controllers. I judged this necessary rather than out-of-scope: acceptance criterion 1 ("GET /management/families/:id renders 200 for an admin") only holds via the actual Show link with this fix in place.

## Tests

Added `spec/requests/management/` (didn't exist before):
- `families_spec.rb` — covers `management/families#index` and `#show`, per the acceptance criteria.
- `friendly_id_show_pages_spec.rb` — covers the friendly_id lookup fix on the other 6 affected resources.

Full suite green (1101 examples, 0 failures), rubocop clean.

## Found but deliberately not fixed here — recommend a follow-up issue

Most `management/*` CRUD controllers' `create`/`update`/`destroy` actions use bare `redirect_to @model` / `redirect_to model_url` (e.g. `redirect_to @genuse`, `redirect_to families_url`). This raises the same kind of `NoMethodError` as the bug this issue is about, because Rails doesn't auto-namespace polymorphic redirects — confirmed by reproducing it against `genuses#create` (`undefined method 'genus_url'`). Affects roughly a dozen controllers (families, divisions, division_classes, division_orders, foreign_sources, foreign_sources_plants, genuses, kingdoms, major_groups, plants, subkingdoms, species_images, users) — i.e. creating, editing or deleting most back-office records is currently broken. `species_controller` is the only one that already does this right (`redirect_to [:management, @species]`). This is a write-path bug across many controllers and deserves its own issue rather than riding along with this show-page fix.

Touches: app/views/management/, app/controllers/management/, spec/requests/management/

### DIFF
--- a/app/controllers/management/division_classes_controller.rb
+++ b/app/controllers/management/division_classes_controller.rb
@@ -63,7 +63,7 @@ class Management::DivisionClassesController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_division_class
-    @division_class = DivisionClass.find(params[:id])
+    @division_class = DivisionClass.friendly.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
--- a/app/controllers/management/division_orders_controller.rb
+++ b/app/controllers/management/division_orders_controller.rb
@@ -63,7 +63,7 @@ class Management::DivisionOrdersController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_division_order
-    @division_order = DivisionOrder.find(params[:id])
+    @division_order = DivisionOrder.friendly.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
--- a/app/controllers/management/divisions_controller.rb
+++ b/app/controllers/management/divisions_controller.rb
@@ -63,7 +63,7 @@ class Management::DivisionsController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_division
-    @division = Division.find(params[:id])
+    @division = Division.friendly.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
--- a/app/controllers/management/families_controller.rb
+++ b/app/controllers/management/families_controller.rb
@@ -63,7 +63,7 @@ class Management::FamiliesController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_family
-    @family = Family.find(params[:id])
+    @family = Family.friendly.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
--- a/app/controllers/management/foreign_sources_controller.rb
+++ b/app/controllers/management/foreign_sources_controller.rb
@@ -64,7 +64,7 @@ class Management::ForeignSourcesController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_foreign_source
-    @foreign_source = ForeignSource.find(params[:id])
+    @foreign_source = ForeignSource.friendly.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
--- a/app/controllers/management/kingdoms_controller.rb
+++ b/app/controllers/management/kingdoms_controller.rb
@@ -63,7 +63,7 @@ class Management::KingdomsController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_kingdom
-    @kingdom = Kingdom.find(params[:id])
+    @kingdom = Kingdom.friendly.find(params[:id])
   end
 
   # Never trust parameters from the scary internet, only allow the white list through.
--- a/app/controllers/management/subkingdoms_controller.rb
+++ b/app/controllers/management/subkingdoms_controller.rb
@@ -63,7 +63,7 @@ class Management::SubkingdomsController < Management::ManagementController
 
   # Use callbacks to share common setup or constraints between actions.
   def set_subkingdom
-    @subkingdom = Subkingdom.find(params[:id])
+    @subkingdom = Subkingdom.friendly.find(params[:id])
   end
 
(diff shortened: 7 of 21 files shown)


lambda2:
🔎 claim-1788867168-26971

lambda2:
## Review: Pass

**CI**: green on head commit `edf033c` (`gh pr checks 338` — Brakeman, RSpec, RuboCop all pass).

**Contract** (#332):
- `families_path` → `management_families_path` fix confirmed at the exact reported line; `config/routes.rb` verified — `resources :families` only exists nested under `api/v1` and `management`, no top-level helper.
- Audit claim verified directly, not taken on faith: grepped every `_path`/`_url` call in every `app/views/management/*/show.html.erb` (18 dirs with a show view) post-fix — all now resolve through `management_*` helpers. `genuses`, `plants`, `species`, `users` show views were already correct, matching the PR's claim.
- Friendly_id lookup fix verified against the actual model list (`grep friendly_id app/models/*.rb`): exactly 7 models are slugged and had `Model.find` in their management controller (`division`, `division_class`, `division_order`, `family`, `foreign_source`, `kingdom`, `subkingdom`) — all 7 switched to `.friendly.find`. Un-slugged models used elsewhere in the same audit (`foreign_sources_plant`, `major_group`, `species_image`, `user`) were correctly left on plain `.find`.
- Noted but not a contract gap: `app/views/management/sessions/show.html.erb` and `species_proposals/show.html.erb` were also fixed, but `Management::SessionsController`/`Management::SpeciesProposalsController` don't exist anywhere in git history — those routes 500 on `uninitialized constant` regardless of this PR's changes. Pre-existing, unrelated to the path-helper bug, out of scope for #332.
- Two new spec files under `spec/requests/management/` (didn't exist before) cover both fixes directly, matching acceptance criteria 1 and 2.
- `Touches:` line matches the diff exactly (`app/views/management/`, `app/controllers/management/`, `spec/requests/management/`) — nothing outside stated scope.

**Local verification** (temp worktree at head commit, `yarn build` run first — without compiled management assets every `/management/*` page 500s, an artifact of a fresh worktree, not the PR):
- `bundle exec brakeman -q` → 0 warnings.
- Targeted specs (`families_spec.rb`, `friendly_id_show_pages_spec.rb`) → 9 examples, 0 failures.
- Full suite → 1101 examples, 0 failures — matches the PR's claimed count exactly. (First attempt hit the known shared-`trefle_test`-database race from #333's thread — a concurrent session's deadlock/pending-migration, not this PR; retried clean.)

**Judgment call, and why it holds up**: the PR explicitly declines to fix the sibling bug it found (unnamespaced `redirect_to @model` in ~12 controllers' create/update/destroy) and recommends a follow-up issue instead, with a reproduction (`genuses#create` → `undefined method 'genus_url'`) and the affected-controller list. That's the right scope boundary — #332's acceptance criteria are only about the show-page GET path, and this write-path bug is a separate, larger fix better tracked on its own.

No gaps found against the acceptance criteria. Converting back to ready for merge-train.

PR #343, opened 2026-09-08
Donate: stop promising an upgrade within an hour, and drop the bold in the sponsor section

### LINKED ISSUE #340: Donate: stop promising an upgrade within an hour, and drop the bold in the sponsor section
The sponsor section of `/donate` makes a promise about timing we should not make, and leans on bold for three things that do not need it.

## 1. Drop the "within an hour" promise

`app/views/home/donate.html.erb:57`

```erb
<p>Your account will then be upgraded within an hour, and your API token will change.</p>
```

Nothing enforces that hour — the upgrade rides on the GitHub sponsors sync job, so the delay depends on when that job next runs and on GitHub's own propagation. Stating a deadline we do not control invites support requests when it is missed. Say that the upgrade follows automatically and that the API token changes, without quantifying the delay.

This is the only occurrence in the codebase: `git grep "within an hour"` returns this line alone. The issue title says "on every page" but the audit found just the one.

## 2. Drop the bold

`app/views/home/donate.html.erb:49`

```erb
<p><b>Sponsors have an extended request limit of 600 requests/minute.</b></p>
```

Bold on a whole sentence is shouting, and the two GitHub steps below it read as emphasised too because they are the only links in the block. The section is short enough to carry its own hierarchy: the heading says what it is, the ordered list says what to do. Set the limit sentence as normal body copy.

## 3. While in there: the markup is invalid

Lines 50-56 nest an `<ol>` inside a `<p>`. The HTML parser closes the paragraph before the list, so the "To do this, you need to" text and the list end up as siblings, and the trailing `</p>` is dropped. Not a visible bug today, but it makes the spacing here behave differently from every other section — worth correcting while the file is open.

## Acceptance criteria

- `/donate` no longer states any upgrade delay; it still says the API token changes.
- `git grep "within an hour"` returns nothing.
- No `<b>` or `<strong>` in the sponsor section.
- The list is a sibling of the paragraph, not nested inside it; the page validates.
- Request specs for `/donate` green.

Touches: `app/views/home/donate.html.erb`
Blocked by: none

lambda2:
Closes #340

- Reword the sponsor upgrade timing claim to not quantify a delay we don't control ("upgraded automatically" instead of "within an hour"), while still stating the API token changes.
- Remove `<b>` emphasis on the request-limit sentence.
- Un-nest the `<ol>` from the enclosing `<p>` — the old markup was invalid (a block element inside a paragraph); the list is now a sibling.

Verified:
- `git grep "within an hour"` returns nothing.
- No `<b>`/`<strong>` remains in the sponsor section.
- Full RSpec suite: 1129 examples, 0 failures.
- RuboCop: 0 offenses.

Touches: `app/views/home/donate.html.erb`

### DIFF
--- a/app/views/home/donate.html.erb
+++ b/app/views/home/donate.html.erb
@@ -47,14 +47,13 @@
         <h1 class="title is-4 ">
           <%= fa_icon('hand-heart', class: 'has-text-success') %> How to sponsor us
         </h1>
-        <p><b>Sponsors have an extended request limit of 600 requests/minute.</b></p>
-        <p>To do this, you need to
-          <ol>
-            <li><a href="<%= profile_path %>" target="_blank">Link your account to Github</a></li>
-            <li><a href="https://github.com/sponsors/treflehq" target="_blank">Sponsor us on Github</a>.</li>
-          </ol>
-        </p>
-        <p>Your account will then be upgraded within an hour, and your API token will change.</p>
+        <p>Sponsors have an extended request limit of 600 requests/minute.</p>
+        <p>To do this, you need to:</p>
+        <ol>
+          <li><a href="<%= profile_path %>" target="_blank">Link your account to Github</a></li>
+          <li><a href="https://github.com/sponsors/treflehq" target="_blank">Sponsor us on Github</a>.</li>
+        </ol>
+        <p>Your account will then be upgraded automatically, and your API token will change.</p>
       </section>
 
 


lambda2:
🔎 claim-1788882249-29397

lambda2:
## Review: Pass

**Checked:** CI (Brakeman/RSpec/RuboCop all green, settled not pending), diff against #340's acceptance criteria, and re-ran the relevant checks locally in the PR's worktree.

All four criteria hold:
- `git grep -i "within an hour"` — no matches (confirmed locally, exit 1).
- No `<b>`/`<strong>` in the sponsor section (confirmed locally, exit 1).
- `<ol>` is now a sibling of the `<p>` tags, not nested inside one — markup validates, matches the pattern used elsewhere in the file.
- `/donate` still states the API token changes, just without quantifying the delay ("upgraded automatically, and your API token will change").
- `spec/requests/web/home_spec.rb` (the `/donate` specs): 16 examples, 0 failures, run locally against this branch.
- `bundle exec rubocop`: 454 files inspected, no offenses.

Diff is scoped exactly to `app/views/home/donate.html.erb` as declared, no scope creep.

No gaps found.

Issue #318, opened 2026-09-07
Expose each upstream source's licence in the API (sources[].licence)


lambda2:
Every species payload already carries a `sources` array (via `ForeignSourcesPlantSerializer`), but it only gives the source name, external id and URL. A consumer who wants to comply with "data keeps the licence of its source" has no way to know what that licence *is*, record by record. This issue makes the upstream licence a first-class, machine-readable part of the payload — it is the foundation the rest of the attribution work (meta.citation #319, /citation page) builds on.

Upstream licences were verified on 2026-09-07:

| Source | Licence | Notes |
|---|---|---|
| POWO / WCVP | CC-BY-4.0 | per the WCVP dataset releases |
| GBIF (backbone) | CC-BY-4.0 | |
| WFO | CC0-1.0 | |
| IPNI | CC-BY-4.0 | to be confirmed with Kew — seed it as NULL until confirmed |
| USDA | Public domain (US Gov) | |
| PlantNet | unclear | seed NULL, clarification underway |
| others (pfaf, flora_*, …) | unclear | seed NULL |

## Scope

1. Migration: add `licence` (string, nullable) and `licence_url` (string, nullable) to `foreign_sources`. SPDX identifiers where one exists (`CC-BY-4.0`, `CC0-1.0`), free text otherwise.
2. Data migration (or migrator worker, consistent with `app/workers/migrators/` conventions) seeding the verified values above. Unverified sources stay NULL — never guess a licence.
3. Expose `licence` and `licence_url` in `ForeignSourcesPlantSerializer` (delegated from the source), so `sources[]` entries become self-describing.
4. Update the JSON schemas in `lib/schemas/` and the request specs that validate against them; swagger/rswag if the sources block is described there.

## Acceptance criteria

- `GET /api/v1/species/:id` shows `licence: "CC-BY-4.0"` on a POWO-linked source entry and `licence: null` on an unverified one.
- Schema validation specs green; no N+1 introduced (the serializer already includes `:foreign_source`).
- RuboCop and Brakeman green.

Touches: db/migrate, app/models/foreign_source.rb, app/serializers/foreign_sources_plant_serializer.rb, app/workers/migrators, lib/schemas, spec
Blocked by: none.
Related: #285 (names Trefle's own licence — independent of this, which only surfaces upstream facts), #21.

lambda2:
🔒 claim-1788774714-22977

PR #362, opened 2026-09-09
Seed the WFO licence, and report the slugs that find no row

### LINKED ISSUE #348: Source licences: WFO is slugged WFO, the migrator looks for wfo, and the spec creates the row it looks up
`Migrators::SourceLicences` matches sources by exact slug. WFO's row in production is slugged `WFO`, the mapping key is `'wfo'`, so `find_by` misses and `next unless fs` drops it without a word. WFO ships with no licence.

## Observed in production

Ran `Migrators::SourceLicencesWorker` on 2026-09-09 right after the v2.5.0 rollout (`jid=ce6928c8f085e7a71172288d`). Three of the four mapped sources were seeded:

```
powo  -> CC-BY-4.0                 OK
gbif  -> CC-BY-4.0                 OK
usda  -> Public domain (US Gov)    OK
wfo   -> CC0-1.0                   NOT APPLIED
```

The full `foreign_sources` table, 17 rows, with the casing as stored:

```
catminat                  flora_iberica    flora_of_north_america
gbif        <- seeded     IPNI             merged_from_synonym
missouri_botanical_garden openfarm         pfaf
plantnet                  powo <- seeded   tela-botanica
treesandshrubs            tropicos         usda <- seeded
WFO         <- MISSED     wikipedia
```

`ForeignSource.find_by(slug: 'wfo')` returns nil; `find_by(slug: 'WFO')` returns the row. Two of the seventeen slugs are uppercase — `WFO` and `IPNI` — so the casing in this table is not consistent, and any future mapping entry for IPNI would fail the same way.

This is user-visible: WFO is a primary taxonomic source and appears first in `sources[]` on species payloads. Verified on production, the API now returns

```json
{"id": "wfo-0000000003", "name": "WFO", "licence": null, "licence_url": null}
```

while GBIF on the same endpoint correctly returns `"licence": "CC-BY-4.0"`. Naming each source's licence is the entire point of #318/#324, so this leaves that work half-delivered for one of the biggest sources.

## Why it was not caught

Two reasons, both worth fixing.

**1. The failure is silent.** `lib/migrators/source_licences.rb`:

```ruby
LICENCES.each do |slug, attrs|
  fs = ForeignSource.find_by(slug: slug)

  next unless fs      # <- no log, no counter, no raise

  fs.update!(attrs)
end
```

A run that seeds 3 of 4 is indistinguishable from a run that seeds 4 of 4. Nothing in the output says otherwise.

**2. The spec creates the row it is about to look up.** `spec/lib/migrators_spec.rb:123`:

```ruby
it 'seeds a source that is not part of the fixed seed set (WFO)' do
  wfo = ForeignSource.create!(name: 'World Flora Online', slug: 'wfo')
  described_class.run
  expect(wfo.reload.licence).to eq('CC0-1.0')
end
```

It builds the fixture with the slug the code expects, so it asserts the mapping against itself and can never detect a mismatch with the real row. The test is green while production is wrong.

The migrator's header comment shows the author reasoned about WFO being absent, but landed on the wrong explanation:

> `db/botanic_seeds.rb` (the tracked seed data) has no WFO entry, so if production hasn't crawled a WFO-linked species yet, that row doesn't exist and this run no-ops for it — safe to re-run once it does.

The row does exist and has since at least 2022 (`last_update` on its species links). Re-running will never fix it, because the reason is the casing, not the absence.

## Scope

- Make the lookup case-insensitive (`where('lower(slug) = ?', slug.downcase).first` or equivalent), or key the mapping on the stored casing. Case-insensitive is preferable: it also covers `IPNI` and anything else added later.
- Make a miss visible. At minimum log which mapped slugs found no row and how many were seeded; a migrator that silently half-runs is the actual defect here. Consider making `Migrators::SourceLicences.run` return or log a summary the way the other migrators report.
- Fix the spec so it exercises the real casing: assert against a row created as `WFO`, and add a regression case for a slug whose casing differs from the mapping key. The current WFO example should not be able to pass if the lookup is case-sensitive.
- Re-run the worker on production afterwards and confirm WFO carries `CC0-1.0`.

Deliberately **not** in scope: filling in the other 13 sources. The migrator's policy of only seeding confirmed licences and never guessing is correct and should stay.

## Acceptance criteria

- `Migrators::SourceLicences.run` seeds WFO with `CC0-1.0` and `https://creativecommons.org/publicdomain/zero/1.0/` against a row slugged `WFO`.
- A slug present in `LICENCES` but absent from the table is reported, not silently skipped.
- Sources outside the mapping still stay NULL — the existing "never guess" specs keep passing.
- A regression spec fails if the lookup is made case-sensitive again.
- After the deploy, `ForeignSource.where.not(licence: nil).count` is 4 in production, and `sources[]` on a WFO-linked species shows the licence.

Touches: `lib/migrators/source_licences.rb`, `spec/lib/migrators_spec.rb`
Blocked by: none

lambda2:
Closes #348.

Ran `Migrators::SourceLicencesWorker` on production on 2026-09-09, right after the 2.5.0 rollout. It seeded three of the four mapped sources and reported nothing about the fourth:

```
powo  -> CC-BY-4.0                OK
gbif  -> CC-BY-4.0                OK
usda  -> Public domain (US Gov)   OK
wfo   -> CC0-1.0                  NOT APPLIED
```

WFO is stored as `WFO`; the mapping key is `wfo`; `find_by(slug:)` missed it and `next unless fs` dropped it silently. Two of the seventeen slugs are upper-case — `WFO` and `IPNI` — so the same trap was waiting for IPNI the day its licence is confirmed.

This is user-visible. On production the API returned:

```json
{"id": "wfo-0000000003", "name": "WFO", "licence": null, "licence_url": null}
```

while GBIF on the same payload carried `"licence": "CC-BY-4.0"`. Naming each source's licence is the point of #318/#324, and it was half-delivered on the largest taxonomic source.

## Change

- Case-insensitive slug match.
- `run` returns `{seeded:, missing:}` and logs the misses. A run that seeds 3 of 4 was previously indistinguishable from one that seeded 4 of 4.
- The header comment carried the wrong diagnosis — it reasoned that WFO might not exist yet because `db/botanic_seeds.rb` has no entry for it. The row does exist, and re-running would never have fixed it. Corrected.

## The spec could not have caught it

```ruby
wfo = ForeignSource.create!(name: 'World Flora Online', slug: 'wfo')
described_class.run
expect(wfo.reload.licence).to eq('CC0-1.0')
```

It created the row with the exact string the mapping looks up, so it asserted the mapping against itself and stayed green while production was wrong. It now builds `WFO` and fails if the lookup is made case-sensitive again. A second example covers the reporting path.

Both new examples fail on master and pass here:

```
without the fix: 15 examples, 2 failures
with the fix:    15 examples, 0 failures
full suite:      1137 examples, 0 failures
```

## After merge

The worker needs re-running on production to seed WFO. It is idempotent.

Deliberately unchanged: the policy of seeding only confirmed licences and never guessing. The other thirteen sources stay NULL.

Touches: `lib/migrators/source_licences.rb`, `spec/lib/migrators_spec.rb`

### DIFF
--- a/lib/migrators/source_licences.rb
+++ b/lib/migrators/source_licences.rb
@@ -2,10 +2,9 @@
 # 2026-09-07. Only sources with a confirmed licence are seeded here — every
 # other source (IPNI, PlantNet, pfaf, flora_*...) stays NULL, never guessed.
 #
-# Matches by slug only, and never creates a ForeignSource: `db/botanic_seeds.rb`
-# (the tracked seed data) has no WFO entry, so if production hasn't crawled a
-# WFO-linked species yet, that row doesn't exist and this run no-ops for it —
-# safe to re-run once it does.
+# Never creates a ForeignSource: it seeds a licence onto a row that already
+# exists, and reports the slugs it could not find instead of passing over them.
+# Matching is case-insensitive because the stored casing is not uniform.
 module Migrators
   class SourceLicences
 
@@ -27,14 +26,30 @@ module Migrators
       }
     }.freeze
 
+    # Slug casing is not consistent in foreign_sources -- WFO and IPNI are
+    # stored upper-case, everything else lower-case -- so an exact match
+    # silently skipped WFO and left the biggest taxonomic source without its
+    # licence in production. Match case-insensitively, and say what was missed
+    # rather than returning as if all four had been seeded.
     def self.run
-      LICENCES.each do |slug, attrs|
-        fs = ForeignSource.find_by(slug: slug)
+      seeded = []
+      missing = []
 
-        next unless fs
+      LICENCES.each do |slug, attrs|
+        fs = ForeignSource.where('lower(slug) = ?', slug.downcase).first
+
+        if fs.nil?
+          missing << slug
+          next
+        end
 
         fs.update!(attrs)
+        seeded << fs.slug
       end
+
+      Rails.logger.warn("[SourceLicences] no foreign_source for #{missing.join(', ')}") if missing.any?
+
+      { seeded: seeded, missing: missing }
     end
 
   end
--- a/spec/lib/migrators_spec.rb
+++ b/spec/lib/migrators_spec.rb
@@ -120,12 +120,26 @@ RSpec.describe Migrators do
       expect(powo.reload.licence_url).to eq('https://creativecommons.org/licenses/by/4.0/')
     end
 
-    it 'seeds a source that is not part of the fixed seed set (WFO)' do
-      wfo = ForeignSource.create!(name: 'World Flora Online', slug: 'wfo')
+    it 'seeds WFO, whose slug is stored upper-case in production' do
+      # Built as 'WFO' on purpose: the previous version of this example created
+      # it as 'wfo', the exact string the mapping looks up, so it asserted the
+      # mapping against itself and stayed green while production had no licence
+      # on its largest taxonomic source.
+      wfo = ForeignSource.create!(name: 'World Flora Online', slug: 'WFO')
 
       described_class.run
 
       expect(wfo.reload.licence).to eq('CC0-1.0')
+      expect(wfo.reload.licence_url).to eq('https://creativecommons.org/publicdomain/zero/1.0/')
+    end
+
+    it 'reports a mapped slug that has no row rather than skipping it silently' do
+      ForeignSource.where('lower(slug) = ?', 'usda').delete_all
+
+      result = described_class.run
+
+      expect(result[:missing]).to include('usda')
+      expect(result[:seeded]).not_to include('usda')
     end
 
     it 'leaves an unconfirmed source at NULL rather than guess' do


lambda2:
🔎 claim: reviewing this PR now.

lambda2 (commented):
## Review: Pass

Checked against #348's acceptance criteria:

- **WFO seeded against the real casing** — new spec builds the fixture as `ForeignSource.create!(slug: 'WFO')` (matching production), and `Migrators::SourceLicences.run` now matches via `where('lower(slug) = ?', slug.downcase)`. This is the crux of the bug: the old spec created its fixture as `slug: 'wfo'`, asserting the mapping against itself. The new one would fail if case-sensitivity crept back in.
- **Misses are reported, not swallowed** — `run` now returns `{seeded:, missing:}` and logs `Rails.logger.warn` when `missing.any?`. New spec deletes the `usda` row and asserts it lands in `missing`, not `seeded`.
- **"Never guess" policy intact** — the two pre-existing specs (`leaves an unconfirmed source at NULL`, `is a no-op for a source outside the verified set`) are untouched and still pass.
- **Scope** — diff is exactly the two files in `Touches:` (`lib/migrators/source_licences.rb`, `spec/lib/migrators_spec.rb`). The header comment's corrected diagnosis is a welcome side effect, not scope creep.

**CI**: RSpec/RuboCop/Brakeman all `pass` (run 34407123366).

**Local verification** (worktree on `fix-wfo-licence`, commit `6024934`):
```
bundle exec rspec spec/lib/migrators_spec.rb
15 examples, 0 failures

bundle exec rubocop lib/migrators/source_licences.rb spec/lib/migrators_spec.rb
2 files inspected, no offenses detected
```

Note for after merge (per PR body, not blocking): the worker needs re-running on production to actually seed WFO — this PR only fixes the code.

PR #368, opened 2026-09-14
Turn three client-triggerable Rack-level 500s into 400s

### LINKED ISSUE #331: Three Rack-level 500s on malformed requests: IP spoof check, rack-attack token parsing, method override
Three distinct exception classes reach Sentry from the same place: a malformed request that never gets near a controller, and the app answers `500`. All three are ours to fix in the Rack layer, and two of them are in `config/initializers/rack_attack.rb`.

Sentry: [API-82](https://trefle.sentry.io/issues/API-82) (176 events), [API-89](https://trefle.sentry.io/issues/API-89) (53 events, **escalating**), [API-8C](https://trefle.sentry.io/issues/API-8C) (4 events). Release `v2.5.1`, environment `production`.

## A note on the Sentry culprit

API-82 and API-8C are both filed under `RateLimitHeadersMiddleware in call`, which is misleading. That middleware is `unshift`ed to the very top of the stack (`config/initializers/rack_attack.rb:93`), so every unhandled downstream exception surfaces with its frame as the culprit. Nothing is wrong inside it. Read the second-to-last frame instead.

## 1. `IpSpoofAttackError` — 176 events (API-82)

```
ActionDispatch::RemoteIp::IpSpoofAttackError: IP spoofing attack?!
  HTTP_CLIENT_IP="127.0.0.1" HTTP_X_FORWARDED_FOR="10.42.0.1"
```

Raised from `Rails::Rack::Logger#call_app` → `request.remote_ip`, i.e. **while logging the request line**, before routing. Any request carrying a `Client-IP` header that disagrees with `X-Forwarded-For` 500s, whatever the path — the sampled event is `POST /blog/wordpress/wp-json/batch/v1` from a WordPress scanner.

`X-Forwarded-For: 10.42.0.1` is the cluster's own ingress address. The app has **no** `trusted_proxies` configuration anywhere (`grep -rn "trusted_proxies\|ip_spoofing" config/` is empty), so `RemoteIp` treats the ingress as untrusted and compares two client-controlled headers. A header that any client can set must not be able to produce a 500.

## 2. `Rack::Multipart::EmptyContentError` — 53 events, escalating (API-89)

Root cause is our own safelist, `config/initializers/rack_attack.rb:8`:

```ruby
def self.token(req)
  [
    req.env['HTTP_AUTHORIZATION'].to_s.downcase.gsub(/bearer /, ''),
    req.params['token'].to_s.downcase.gsub(/\s+/, '')   # <- here
  ].reject(&:blank?).first
end
```

`req.params` is `GET.merge(POST)`, so reading it **forces a full parse of the request body** for every request rack-attack evaluates — including `POST /signin`. A malformed multipart body raises out of `Rack::Request#POST`, through `Rack::Attack::Check#matched_by?`, and out of the whole stack as a 500 (`http.response.status_code: 500` is on the event).

Two things are wrong: a rate-limit safelist should not parse request bodies at all (it is work done before throttling, on unthrottled traffic), and it must not be able to raise.

## 3. `Encoding::CompatibilityError` — 4 events (API-8C)

```
Encoding::CompatibilityError: incompatible character encodings: UTF-16LE and UTF-8
```

From `Rack::MethodOverride#method_override_param` → `req.POST` → `Rack::QueryParser#_normalize_params` → `name.index('[', 1)`, on a form field name submitted as UTF-16LE. Same shape as #2: body parsing that a client can make raise. Fixing #2 does not fix this one — `Rack::MethodOverride` parses the body on its own.

## Scope

- Configure `RemoteIp` for the real topology: set `config.action_dispatch.trusted_proxies` to include the cluster's internal ranges so the ingress hop is trusted. Do not simply set `ip_spoofing_check = false` without saying why — `request.remote_ip` feeds rack-attack throttling and `Api::ApiController#check_jwt!`, so silently trusting a client-supplied `Client-IP` would let a caller pick their own throttle bucket. Say in the PR which ranges are trusted and what that means for throttling.
- Stop `Rack::Attack.token` from parsing bodies: read the token from `req.env['HTTP_AUTHORIZATION']` and the **query string** (`req.GET`) only. A token in a POST body was never a documented way to authenticate (`POST /api/auth/claim` takes JSON), so confirm against `spec/requests` before changing it, and keep the behaviour if a spec depends on it — in that case wrap the parse instead.
- Make the edge fail as `400`, not `500`: a malformed request body or header is a client error. The outermost middleware is the natural place; it already wraps everything. Return the documented `/api` error envelope for `/api` paths.
- These are unauthenticated, externally triggerable 500s. Whatever the fix, none of the three should be reachable from a header or body a client controls.

## Acceptance criteria

- A request with `Client-IP: 127.0.0.1` behind the ingress does not raise; `request.remote_ip` resolves to the real client address.
- `POST /signin` with a truncated/empty multipart body returns `400`, not `500`.
- A POST whose form field names are not valid UTF-8 returns `400`, not `500`.
- Specs cover all three at the Rack level (request specs with the offending headers/bodies), each failing against current `master`.
- Rate limiting still buckets by the true client IP — a spec pins that a client-supplied `Client-IP` cannot change the throttle key.

Sentry: `Fixes API-82`, `Fixes API-89`, `Fixes API-8C`.

Touches: `config/initializers/rack_attack.rb`, `config/environments/production.rb`, `config/application.rb`, `spec/requests`
Blocked by: none

lambda2:
Closes #331

## What
Three exception classes could reach production as an unhandled 500 from nothing but a malformed header or body, before any controller ran: `ActionDispatch::RemoteIp::IpSpoofAttackError` (API-82), `Rack::Multipart::EmptyContentError` (API-89), `Encoding::CompatibilityError` (API-8C).

## Changes
- **`config/initializers/rack_attack.rb`**: `Rack::Attack.token` now reads only `req.GET`, never `req.params` (`GET.merge(POST)`). The safelist block runs on *every* request regardless of path, so it was forcing a full body parse before any throttling happened — that's what let a broken multipart body 500 out of `Rack::Attack` itself, independent of the controller. A token was never documented as accepted from a POST body (the `token` security scheme is `apiKey, in: query`), so no supported behaviour changes — confirmed against `spec/requests` (900_auth_spec passes `token` via query in all cases).
- **`config/application.rb`**: `config.action_dispatch.trusted_proxies` is now explicit, extending Rails' default `TRUSTED_PROXIES` with the cluster's ingress network (`10.42.0.0/16`, inferred from the reported event's `X-Forwarded-For: 10.42.0.1`). Rails' default already covers RFC1918 space, so this rarely changes behaviour in practice — it's here so the trust boundary is documented rather than implicit, and so request specs can exercise it deterministically. Set in `application.rb` (not `production.rb`, despite the issue's Touches list) so the test environment shares it.
- **`config/initializers/rack_attack.rb`** — `RateLimitHeadersMiddleware`, already `unshift`ed to the very top of the middleware stack (the issue's own note on the misleading Sentry culprit explains why), now also downgrades all three exception classes to 400 — the documented `/api` envelope on `/api` paths, plain text elsewhere. It catches them two ways, because they don't all reach it the same way: `IpSpoofAttackError` and `Encoding::CompatibilityError` (from `Rack::MethodOverride`) escape unrescued all the way up, so a plain `rescue` catches them; `EmptyContentError` (from `Rack::Attack`'s own safelist) gets rescued *deeper* in the stack by `ActionDispatch::ShowExceptions` into a generic 500 first, so this reads `env['action_dispatch.exception']` (set by ShowExceptions before it renders) to catch that case too.

## A deliberate deviation from the literal acceptance criterion
The issue's first acceptance criterion says a conflicting `Client-IP`/`X-Forwarded-For` request "does not raise; `request.remote_ip` resolves to the real client address." Checked against the installed Rails source: `ActionDispatch::RemoteIp::GetIp#calculate_ip` raises `IpSpoofAttackError` whenever both headers are present and disagree, **regardless of `trusted_proxies`** — that list only affects which IP wins *after* the spoof check passes. So `trusted_proxies` cannot make this specific request "resolve" to anything; the two headers are genuinely contradictory, and there's no way to know which one is real. What this PR does instead: reject it with 400 (via the `RateLimitHeadersMiddleware` rescue), which satisfies the actual goal (no unauthenticated externally-triggerable 500) without pretending to resolve an address that doesn't exist for that request. `trusted_proxies` remains valuable for ordinary multi-hop traffic (a single trusted `X-Forwarded-For` hop, no `Client-IP`) and as documentation of the trust boundary.

## Also discovered, left out of scope
`ActionDispatch::RemoteIp` trusts a **lone** `Client-IP` header outright when `X-Forwarded-For` is absent (verified empirically: `request.remote_ip` resolves directly to whatever `Client-IP` claims). This is stock Rails behaviour, not something this issue introduced, but it does mean `Api::ApiController#check_jwt!`'s IP-lock can be bypassed by a client that only sets `Client-IP`. Flagging it here since it's adjacent to this fix, but it's a separate, bigger hardening question and not part of #331's scope.

## Specs
`spec/requests/rack_level_malformed_requests_spec.rb`:
- Conflicting `Client-IP`/`X-Forwarded-For` → 400 (was 500).
- A `Content-Length`-accurate but structurally truncated multipart body (opening boundary line, nothing after) → 400 (was 500). A fully empty body doesn't reproduce this — Rack's multipart parser short-circuits before ever raising when `content_length == 0`.
- A multipart part whose own `Content-Type: text/plain; charset=UTF-16LE` header retags its field *name* (not just value) with a non-ASCII-compatible encoding → 400 (was 500). Reproduces the exact reported frame (`Rack::MethodOverride#method_override_param` → `req.POST` → `Rack::QueryParser#_normalize_params`).
- A regression spec pinning that a client-supplied `Client-IP` cannot change the rate-limit throttle key (`Rack::Attack`'s throttle key uses `Rack::Request#ip`, which only consults `REMOTE_ADDR`/`X-Forwarded-For`, never `Client-IP` — verified empirically, distinct from `ActionDispatch::Request#remote_ip`'s more permissive resolution noted above).

All three malformed-request specs were confirmed to fail (500) against `master` before the fix, and pass after.

## Test status
- `bundle exec rubocop`: 457 files, 0 offenses.
- `bundle exec brakeman -q`: 0 warnings.
- `bundle exec rspec spec/requests`: 284 examples, 75 failures — identical failure set (by name) as on `master` with these changes reverted; this PR introduces zero new failures. Pre-existing baseline, not addressed here.

Touches: `config/application.rb`, `config/initializers/rack_attack.rb`, `spec/requests`


lambda2:
🔎 claim-1789372195-10871

lambda2:
## Review: Fail — needs-fixes

CI is green (RSpec/RuboCop/Brakeman all pass) and the local repro is solid: I checked out the branch in a worktree, confirmed all 3 new malformed-request specs fail (500/raised) against `master` and pass on this branch, and diffed the full `spec/requests` failure set against `master` — zero new failures introduced. The `trusted_proxies` deviation write-up is also correct: I verified against the installed Rails source (`actionpack-8.0.5.1/lib/action_dispatch/middleware/remote_ip.rb`) that `GetIp#calculate_ip` raises `IpSpoofAttackError` whenever `Client-IP`/`X-Forwarded-For` disagree **regardless of the trusted-proxies list** — the 400 fallback for that case is the right call, not a shortcut.

**One acceptance criterion is not actually met, though: "Return the documented `/api` error envelope for `/api` paths" fails for the `EmptyContentError` case specifically.**

- `config/initializers/rack_attack.rb:111-153` (`RateLimitHeadersMiddleware`) and `MALFORMED_REQUEST_ERRORS` (line 112) assume all three exception classes reach this middleware the same way — either as a raised exception (`rescue *MALFORMED_REQUEST_ERRORS`) or via `env['action_dispatch.exception']` (for `Rack::Multipart::EmptyContentError`, per the comment at lines 91-96: *"gets rescued deeper in the stack by `ActionDispatch::ShowExceptions` into a generic 500 first, so this reads `env['action_dispatch.exception']`... to catch that case too"*).
- That's not what happens for a real `/api` endpoint. I reproduced it directly:
  ```ruby
  post '/api/v1/species/1/report', params: "--AaB03x\r\n", headers: { 'CONTENT_TYPE' => 'multipart/form-data; boundary=AaB03x' }
  # => status 400 (correct), but Content-Type: text/html, body is Rails' full HTML
  #    "Action Controller: Exception caught" debug page — NOT the documented
  #    {"error":true,"code":"bad_request",...} envelope. Same result on
  #    /api/auth/claim.
  ```
  Root cause: `Rack::MethodOverride#method_override_param` already has its own `rescue EOFError` (line ~48 of `rack/method_override.rb`), which silently swallows the truncated-multipart parse and lets the request continue — it never reaches `RateLimitHeadersMiddleware` as a raised exception the way `Encoding::CompatibilityError` does. The parse is then retried for real inside the normal Rails request-parameter path, where `ActionDispatch::Request#POST` (`action_dispatch/http/request.rb:432-433`) *already* rescues `EOFError` and re-raises `ActionController::BadRequest` — a class Rails' own `ExceptionWrapper` already maps to 400. By the time anything reaches `env['action_dispatch.exception']`, it's `ActionController::BadRequest`, not `Rack::Multipart::EmptyContentError`, so `malformed_request_error?` never matches and this middleware's handling for this exception class is dead code in every path I tried.
  - For contrast, the other two classes *do* work correctly on `/api` paths — I checked with the same reproduction pattern (conflicting `Client-IP`/`X-Forwarded-For` and the UTF-16LE field-name case against `/api/v1/species/1/report`), both correctly return the JSON `{"error":true,"code":"bad_request",...}` envelope. This is specific to `EmptyContentError`.
- The new spec (`spec/requests/rack_level_malformed_requests_spec.rb:34-39`) doesn't catch this because it (a) only targets `/users/sign_in`, never an `/api/*` route, for this exception class, and (b) only asserts `have_http_status(:bad_request)`, never the response `Content-Type`/body shape.

**What "fixed" looks like:** a truncated/empty multipart `POST` to an `/api/*` route must return the same JSON envelope shape as the other two cases, not Rails' generic HTML error page. Add a request spec against an actual `/api/...` route (not just `/users/sign_in`) for this exception class that asserts on `Content-Type: application/json` and the envelope body, the way the spoof-header and encoding-error cases already implicitly get covered by hitting `/api/v1/`. Once real, also fix the stale comment at `config/initializers/rack_attack.rb:91-96` — it describes a rescue path that isn't actually exercised.

🔎 claim-1789372195-10871

lambda2:
🔧 claim-1789373222-13529

lambda2:
Fixed the acceptance-criterion gap from the review.

Root cause was more specific than the stale comment claimed: by the time a truncated multipart body matters, `Rack::MethodOverride`'s own parse attempt has already swallowed the raw `Rack::Multipart::EmptyContentError` (rescuing the `EOFError` it produces). The *next* thing to touch POST params re-parses the body for real and gets `ActionController::BadRequest` instead — raised deep inside `ActionController::Instrumentation#process_action` while building its log payload, before `super` ever reaches a controller's own `rescue_from`. Whenever the request is considered "local" (always true in dev/test via `config.consider_all_requests_local`), `ActionDispatch::DebugExceptions` catches and renders that itself, returning normally without ever touching `env['action_dispatch.exception']` — so neither existing mechanism in `RateLimitHeadersMiddleware` ever saw it, on `/api` or elsewhere.

Fix: force the same POST parse ourselves in the middleware, before calling into the rest of the stack, scoped to `multipart/form-data` so every other request (the overwhelming majority) still skips a body parse it doesn't need — same reasoning #331 already established for not doing this unconditionally. Also added `ActionController::BadRequest` to the malformed-request class list for the production/non-local path where it *does* reach `env['action_dispatch.exception']`, and fixed the stale comment.

Added a request spec against a real `/api/*` route (`POST /api/auth/claim`) asserting `Content-Type: application/json` and the envelope body, not just the status — that's what let the original bug slip past the existing suite.

RSpec (full suite, 1142 examples), RuboCop, and Brakeman all green locally.

lambda2:
🔎 claim-1789374120-20167

lambda2:

**CI**: RSpec/RuboCop/Brakeman all green at `e67b533` (`gh pr checks 368`), the exact commit that landed the fix for the prior review's finding.

**Local verification** (worktree `.claude/worktrees/issue-331-rack-500s`, pulled to `e67b533`):

- `spec/requests/rack_level_malformed_requests_spec.rb`: 5/5 pass, including the new `/api/auth/claim` envelope spec that was missing when this PR previously failed review.
- Reproduced the fix directly: reverted just `config/application.rb` + `config/initializers/rack_attack.rb` to the pre-fix commit (`8b78aae`), keeping the new spec — the envelope spec fails exactly as the last review described (`expected: "application/json", got: "text/html; charset=UTF-8"`). Restored to `e67b533`, clean again.
- Full `spec/requests`: 285 examples, 0 failures. `bundle exec rubocop` on the three touched files: 0 offenses.
- Traced the actual fix logic rather than trusting the commit message: `RateLimitHeadersMiddleware#call` now does `request.POST if request.media_type == 'multipart/form-data'` before calling `@app.call(env)`, scoped so only multipart requests pay for the early parse. Confirmed in Rails source (`actionpack-8.0.5.1/lib/action_dispatch/http/request.rb:408-433`) that `ActionDispatch::Request#POST` caches its result via `fetch_header("action_dispatch.request.request_parameters")` on the shared `env` — so this doesn't double-parse the body for legitimate multipart requests downstream, it just moves the same parse earlier and lets the method-level `rescue *MALFORMED_REQUEST_ERRORS` (now including `ActionController::BadRequest`) catch it before `ActionDispatch::DebugExceptions`/`ShowExceptions` ever see it.
- Checked whether the app has any real multipart upload endpoints that this early parse could regress — it doesn't (`grep` for ActiveStorage/Paperclip/attach turns up nothing; Trefle doesn't accept file uploads), so this is a low-risk, well-scoped change.
- Re-confirmed the two previously-verified cases (conflicting `Client-IP`/`X-Forwarded-For`, non-UTF-8 field name) and the rate-limit-key spec still pass unaffected by this change.

All five acceptance criteria from #331 are now met, including the `/api` envelope requirement the prior review flagged as unmet.

🔎 claim-1789374120-20167

PR #369, opened 2026-09-14
Preload foreign_sources_plants on the explore species menu

### LINKED ISSUE #363: N+1 on foreign_sources in the explore species menu: 7,654 events from one partial
One partial is responsible for **7,654 grouped events** across two Sentry issues, and it has been firing since October 2025.

Sentry: [API-52](https://trefle.sentry.io/issues/API-52) — `Explore::SpeciesController#show`, 4,514 events, first seen 2025-10-06 — and [API-53](https://trefle.sentry.io/issues/API-53) — `Explore::RecordCorrectionsController#index`, 3,140 events, first seen 2025-10-06. [API-5D](https://trefle.sentry.io/issues/API-5D) (`Explore::RecordCorrectionsController#show`, 1 event, new) is the same shape on the third route. Both carry the identical query pattern:

```sql
SELECT "foreign_sources".* FROM "foreign_sources" WHERE "foreign_sources"."id" = $1 LIMIT $2
```

## Why two unrelated controllers report the same N+1

They render the same partial. `app/views/explore/record_corrections/index.html.erb:10` renders `explore/species/species_menu`, which the species page renders too, and the partial walks `foreign_sources_plants` one association at a time:

```erb
app/views/explore/species/_species_menu.html.erb:35    <% if species.foreign_sources.any? %>
app/views/explore/species/_species_menu.html.erb:40      <% species.foreign_sources_plants.each do |fsp| %>
app/views/explore/species/_species_menu.html.erb:44        See on <%= "#{fsp.foreign_source&.name}" %>
```

`fsp.foreign_source` is a `belongs_to` with nothing preloaded, so it is one `SELECT` per source per species. Line 35 adds a second full traversal of its own.

The species page has a third instance of the same mistake in its synonyms section:

```erb
app/views/explore/species/show.html.erb:46    <% s.foreign_sources_plants.each do |fo| %>
app/views/explore/species/show.html.erb:47      <small><%= fo.fid %> on <%= fo&.foreign_source&.name %></small>
```

That one is nested inside `@species.synonyms.each`, so it multiplies: synonyms × sources queries on a single page view.

## Where the preload is missing

`Explore::SpeciesController#set_species` (`app/controllers/explore/species_controller.rb:189`) loads a bare record:

```ruby
def set_species
  @species = Species.friendly_or_synonym!(params[:id])
end
```

`Explore::RecordCorrectionsController#index` (line 13) does the same for `@species`, and `#show` reaches it through `@record_correction.record`.

Note that `foreign_sources` is a small, effectively static table — this is a textbook preload, not a query that needs rethinking.

## Scope

- Preload the chain on all three entry points: `foreign_sources_plants: :foreign_source`, plus `synonyms: { foreign_sources_plants: :foreign_source }` for the species page's synonyms section.
- `Explore::RecordCorrectionsController#show` reaches the species through `@record_correction.record` (polymorphic) — preloading there needs care; if it cannot be expressed cleanly, say so in the PR and handle it in the view instead.
- `species.foreign_sources.any?` on line 35 loads a second association to answer a question the already-loaded `foreign_sources_plants` can answer. Collapse it.
- Check whether the same partial is rendered from any other route before declaring the sweep complete (`grep -rn "species_menu" app/views/`).

## Acceptance criteria

- `GET /explore/species/:slug`, `/explore/species/:slug/corrections` and `/explore/corrections/:id` each issue a bounded number of `foreign_sources` queries — one, not one per row.
- A request spec asserts the query count on a species with several synonyms and several sources, and fails against current `master`.
- No behaviour change in the rendered page.

Sentry: `Fixes API-52`, `Fixes API-53`, `Fixes API-5D`.

Touches: `app/controllers/explore/species_controller.rb`, `app/controllers/explore/record_corrections_controller.rb`, `app/views/explore/species/_species_menu.html.erb`, `app/views/explore/species/show.html.erb`, `spec/requests`
Blocked by: none

lambda2:
Closes #363

`explore/species/_species_menu.html.erb` walks `species.foreign_sources_plants.each { fsp.foreign_source }` on every render, and separately calls `species.foreign_sources.any?` to decide whether to show the section at all. Neither association was preloaded on any of the three entry points that render the partial, so the menu cost one query per linked source (doubled by the `.any?` check) — API-52 / API-53 / API-5D, 7,654 grouped Sentry events since October 2025.

## Changes

- `Explore::ExploreController#preload_foreign_sources_for`, called from all three entry points (`SpeciesController#show`, `RecordCorrectionsController#index` and `#show`), preloads `foreign_sources_plants: :foreign_source` in place on the already-loaded record via `ActiveRecord::Associations::Preloader`. A plain `.includes` can't reach these records: `Species.friendly_or_synonym!`'s synonym fallback and `RecordCorrection#record` (polymorphic) both hand back an instance outside of a chainable relation, which is why the preload happens after the record is loaded instead of on the query that loads it.
- For a `Species` record, the same chain is also preloaded on `synonyms`, matching the scope the issue asked for. **Note:** the species page's synonyms `foreign_sources_plants.each` loop (`show.html.erb`) is currently dead code — it's wrapped in a `=begin`/`=end` Ruby comment, the same convention used in 7 other views in this codebase (`grep -rn "=begin" app/views/`) to disable a block. I verified this by compiling the template through Erubi and confirming the loop and its HTML output fall entirely inside the comment. So this preload costs nothing today, but keeps the fix correct if that block is ever re-enabled, and matches the literal scope of #363.
- Collapsed `species.foreign_sources.any?` → `species.foreign_sources_plants.any?` in the partial, so the visibility check reads from the association that's already preloaded instead of loading a second one.

## Testing

`spec/requests/explore/foreign_sources_menu_preload_spec.rb` (new) asserts the query count on `/explore/species/:slug`, `/explore/species/:slug/corrections` and `/explore/corrections/:id` stays flat as synonyms and foreign sources are added — verified to fail against `master` (via `git stash`) before the fix, and to pass after. Two more specs pin that the links section renders correctly and is hidden when a species has no source.

Full RSpec suite (excluding `spec/features`, which needs chromedriver): 1127 examples, 0 failures. RuboCop: 0 offenses repo-wide.


lambda2:
🔎 claim-1789373528-12085

Converting back to draft for the duration of the review (per /review-pr claim step).

lambda2:

**CI**: Brakeman/RSpec/RuboCop all green (`gh pr checks 369`).

**Local verification** (checked out the branch's own worktree, `.claude/worktrees/issue-363-foreign-sources-n1`):

- Full suite excluding `spec/features`: 1127 examples, 0 failures — matches the PR description exactly. RuboCop on the touched files: 0 offenses.
- Confirmed the three new query-count specs actually catch the regression: reverted just the four production files to the pre-fix commit (keeping the new spec), reran — all three query-count specs fail (9→12, 10→13, 11→14 queries) exactly as claimed, then restored the fix and reran clean.
- Verified the sweep is complete: `grep -rn "species_menu" app/views/` finds exactly the three call sites the PR touches (species show, corrections index, corrections show) — no fourth entry point was missed.
- Verified the '=begin/=end dead code' claim independently rather than taking it on faith: compiled `show.html.erb` through Erubi and confirmed `=begin`/`=end` land at column 0 in the generated Ruby source with `s.foreign_sources_plants.each` fully inside the comment block, and that the compiled source still parses (`RubyVM::InstructionSequence.compile`) — the loop is genuinely inert today, so preloading `synonyms: { foreign_sources_plants: :foreign_source }` is optional-but-correct scope-matching, not required for the fix to work.
- Checked `RecordCorrection#record` — polymorphic, can resolve to `Species`, `Synonym`, or `Plant`. But `Explore::RecordCorrectionsController#show` already unconditionally calls `@species.status`/`@species.rank`/`@species.family_name` on the result (pre-existing code, unrelated to this PR), so the controller already assumed Species-only before this change — the new `preload_foreign_sources_for`'s `record.is_a?(Species)` guard doesn't relax or tighten that pre-existing assumption, it's just correctly defensive for the `synonyms` preload leg.
- `species.foreign_sources.any?` → `species.foreign_sources_plants.any?`: confirmed `ForeignSourcesPlant belongs_to :foreign_source` is non-optional, so presence of a `foreign_sources_plants` row implies presence of a `foreign_sources` row — the collapse is behaviorally equivalent, and reads off the already-preloaded association (no added query).
- `ActiveRecord::Associations::Preloader.new(records:, associations:)` keyword signature matches installed Rails 8.0.5.1.

All three acceptance criteria from #363 are met: bounded query count on all three routes (spec-enforced), a request spec that fails against pre-fix code, and no rendered-output change (pinned by the two new content specs).

🔎 claim-1789373528-12085

PR #370, opened 2026-09-14
Load plant.species once and partition by rank instead of six queries

### LINKED ISSUE #364: /api/v1/plants/:id runs six rank queries over the same species set
`GET /api/v1/plants/:id` runs six near-identical queries over the same rows, one per taxonomic rank.

Sentry: [API-8B](https://trefle.sentry.io/issues/API-8B) — `Api::V1::PlantsController#show`, 11 events, first seen 2026-09-10, release `v2.5.1`. [API-5A](https://trefle.sentry.io/issues/API-5A) is the same transaction. Query pattern:

```sql
SELECT "species".* FROM "species" WHERE "species"."plant_id" = $1 AND "species"."rank" = $2
```

## What happens

`PlantSerializer` exposes six sibling collections (`app/serializers/plant_serializer.rb:54-59`):

```ruby
has_many :species_species, serializer: SpeciesLightSerializer, name: :species
has_many :subspecies,      serializer: SpeciesLightSerializer
has_many :varieties,       serializer: SpeciesLightSerializer
has_many :hybrids,         serializer: SpeciesLightSerializer
has_many :forms,           serializer: SpeciesLightSerializer
has_many :subvarieties,    serializer: SpeciesLightSerializer
```

Each is backed by its own scope on `Plant` (`app/models/plant.rb:92-114`):

```ruby
def subspecies
  species.ssp_rank.preload(:plant, :genus, :synonyms)
end
```

Six methods, six scopes, six round trips — all filtering the same `plant.species` set by a different `rank` value. Rendering one plant record therefore costs six queries where one would do.

This is the layer #345 did not reach. Its comment at `app/models/plant.rb:88-91` records that the *inner* preloads were the fix for #281; the six sibling queries themselves were left in place, and Sentry started grouping them as an N+1 two days after that PR merged.

## Scope

- Load `plant.species` **once**, with `preload(:plant, :genus, :synonyms)`, and partition the loaded records by `rank` in Ruby. The six serializer keys stay exactly as they are; only their source changes.
- Keep the ordering each scope currently produces, if any — check the `*_rank` scopes before assuming they are unordered, and pin it in a spec if it matters to the payload.
- `Api::V1::PlantsController#show` (`app/controllers/api/v1/plants_controller.rb:45`) resolves a `Species` and then serializes `@resource.plant`; that is a second place where a preload may belong. Decide where the loading lives (controller vs model) and say why in the PR.
- Do not change the JSON. The six keys, their names and their contents must be byte-identical.

## Acceptance criteria

- `GET /api/v1/plants/:id` issues one query against `species` for the six rank collections, not six.
- A request spec asserts the query count on a plant that has records in several ranks, and fails against current `master`.
- A JSON Schema spec (`lib/schemas/v1/`) still passes: the payload is unchanged.

Sentry: `Fixes API-8B`, `Fixes API-5A`.

Touches: `app/models/plant.rb`, `app/serializers/plant_serializer.rb`, `app/controllers/api/v1/plants_controller.rb`, `spec/requests`
Blocked by: none

lambda2:
Closes #364.

`GET /api/v1/plants/:id` ran six near-identical queries over `species`, one per taxonomic rank (`species_species`, `subspecies`, `varieties`, `hybrids`, `forms`, `subvarieties` on `Plant`), each its own `species.<rank>_rank.preload(...)` scope.

## Change

`Plant` now loads `species.preload(:plant, :genus, :synonyms)` once and partitions the result by `rank` in Ruby, memoized in a private `species_by_rank` method. The six public accessors PlantSerializer calls are unchanged in name and return the same content, just sourced from the single preload.

## Where the preload lives (model vs controller)

Kept in the model. The six rank accessors are only ever called from `PlantSerializer`'s `has_many` declarations, nowhere else in the codebase, so a controller-level preload would have nothing else to serve and the memoized model method already collapses to one query per request. `Api::V1::PlantsController#show` resolves the `Species` via `friendly_or_synonym!` (one query) and reads `.plant` off it (a separate `belongs_to` load) — neither of those needed to change.

## Ordering

The `*_rank` enum scopes carried no explicit `.order`, and `Plant#species` has none either, so there was no ordering guarantee to preserve.

## Testing

- New request spec (`spec/requests/api/v1/plants_rank_collections_spec.rb`) asserts the `species`-table query count on `/api/v1/plants/:id` for a plant with records across several ranks: **7 on `master`, 2 with this fix** (1 to resolve the slug, 1 for the single partitioned preload). Also asserts each rank still comes back under its correct serializer key.
- The existing swagger/JSON-schema spec for `plant` (`spec/requests/api/v1/2_plants_spec.rb`) still passes untouched — payload is byte-identical.
- `bundle exec rubocop app/models/plant.rb spec/requests/api/v1/plants_rank_collections_spec.rb` — 0 offenses.
- Full `bundle exec rspec`: same 90 pre-existing failures on this branch as on unmodified `master` (unrelated feature/web/management specs — worktree asset-build artifacts), no new failures.

Touches: `app/models/plant.rb`, `spec/requests/api/v1`


lambda2:
🔎 claim-1789375104-20485

lambda2:

**CI**: Brakeman/RSpec/RuboCop all green (`gh pr checks 370`).

**Local verification** (worktree `.claude/worktrees/issue-364-plant-rank-preload`, branch pulled to `2a3359c`):

- `spec/requests/api/v1/plants_rank_collections_spec.rb` + `spec/requests/api/v1/2_plants_spec.rb`: 13/13 pass. `bundle exec rubocop app/models/plant.rb spec/requests/api/v1/plants_rank_collections_spec.rb`: 0 offenses.
- Confirmed the new spec actually catches the regression it's meant to guard: reverted only `app/models/plant.rb` to the pre-fix commit (`06361fa`), keeping the new spec — the query-count assertion fails exactly as the PR description claims (`expected: 2, got: 7`). Restored the fix, clean again (2 queries: 1 slug resolve + 1 partitioned `species` preload).
- Confirmed `Species#rank` is a Rails enum (`species/ssp/var/hybrid/form/subvar`) returning string labels, so `species.preload(...).group_by(&:rank)` keys line up exactly with the `species_by_rank['ssp']` etc. lookups — no silent `nil` fallthrough.
- Checked the ordering claim independently: neither the old `*_rank` enum scopes nor `Plant#species` (`has_many :species`, no `default_scope` on `Species`) carry an explicit `.order`, so there was no ordering guarantee to preserve — matches the PR's "Ordering" section.
- Grepped the codebase for other callers of the six accessors (`species_species`, `subspecies`, `varieties`, `hybrids`, `forms`, `subvarieties`) outside `PlantSerializer` and the new spec — none. So the switch from chainable `ActiveRecord::Relation` to a plain `Array` (via `group_by`) can't break a caller that expected further query chaining elsewhere.
- The model vs. controller placement write-up checks out: `Api::V1::PlantsController#show` resolves `Species` via `friendly_or_synonym!` and reads `.plant` off it — unrelated to the six rank accessors, which are only ever invoked from `PlantSerializer`'s `has_many` declarations.

All three acceptance criteria from #364 are met: one `species` query for the six rank collections (spec-enforced and independently reproduced), a request spec that fails against pre-fix code, and the JSON-schema spec for `plant` still passes untouched.

🔎 claim-1789375104-20485

Issue #53, opened 2020-09-22
Ability to add photos


milkintosh:
Some of the primary photos that are used are not the best identifiers, and sometimes there are no photos, so it would be nice to be able to 'submit a correction' for images like you can for other categories.

lambda2:
The corrections pipeline could carry this, but its schema rejects image keys today. Did you mean suggesting an image URL, or uploading a file? The first is a small change, the second needs storage.

PR #371, opened 2026-09-14
Remove ActiveStorage: mounted with no tables, 500ing its upload endpoint

### LINKED ISSUE #365: ActiveStorage is mounted with no tables: its direct-upload endpoint 500s
`ActiveStorage::DirectUploadsController#create` raises on every call because the tables it writes to were never created.

Sentry: [API-8D](https://trefle.sentry.io/issues/API-8D), 5 events, first seen 21 hours ago, release `v2.5.1`.

```
ActiveRecord::StatementInvalid: PG::UndefinedTable:
  ERROR: relation "active_storage_blobs" does not exist
```

## The state of ActiveStorage in this app

The engine is required and configured:

```ruby
config/application.rb:8              require 'active_storage/engine'
config/environments/production.rb:41 config.active_storage.service = :local
config/environments/development.rb:37 config.active_storage.service = :local
config/environments/test.rb:37       config.active_storage.service = :test
```

Requiring the engine mounts its routes, including `POST /rails/active_storage/direct_uploads`, which is what the Sentry events hit.

But the tables do not exist. `db/schema.rb` declares 27 tables and **none** of them are ActiveStorage's — `grep -n active_storage db/schema.rb` is empty. Only the follow-up migrations are present in `db/migrate/`:

```
20200611193330_add_foreign_key_constraint_to_active_storage_attachments_for_blob_id.active_storage.rb
20221206140234_add_service_name_to_active_storage_blobs.active_storage.rb
20221206140235_create_active_storage_variant_records.active_storage.rb
20221206140236_remove_not_null_on_active_storage_blobs_checksum.active_storage.rb
```

The original `create_active_storage_tables` migration is missing, so the amendments have nothing to amend.

And nothing uses it: `grep -rn "has_one_attached\|has_many_attached\|direct_upload" app/` returns nothing. There is no attachment anywhere in the application.

## What this means

The app exposes a public, unauthenticated write endpoint for a feature it does not have, and that endpoint returns `500`. The right fix is almost certainly to remove it, not to create the tables — but that is a product call, because #53 (*Ability to add photos*) is an open request that would eventually want exactly this machinery.

## Scope

- Decide, and say which in the PR: **remove** ActiveStorage (drop the `require`, drop the dead `.active_storage.rb` migrations, drop the three `config.active_storage.service` lines) or **complete** it (generate `active_storage:install`, so the existing amendment migrations apply cleanly).
- Removal is the recommended default: nothing references it, and reinstating it when #53 is actually built is a single generator invocation. Do not keep a broken endpoint live to reserve a feature nobody has started.
- If removing, check that no route outside the engine depends on it and that the `:test` service config is not load-bearing for any spec.
- Either way, `POST /rails/active_storage/direct_uploads` must stop returning `500` — `404` if removed, `2xx`/`4xx` if completed.

## Acceptance criteria

- `POST /rails/active_storage/direct_uploads` no longer returns `500`, pinned by a request spec.
- `bundle exec rails db:migrate` from scratch succeeds and leaves the schema consistent with what the app actually uses.
- The PR states the decision and links #53 so the choice is discoverable when photo upload is picked up.

Sentry: `Fixes API-8D`.

Touches: `config/application.rb`, `config/environments/production.rb`, `config/environments/development.rb`, `config/environments/test.rb`, `db/migrate/`, `db/schema.rb`, `spec/requests`
Blocked by: none

lambda2:
Closes #365.

## Decision

**Removed** ActiveStorage rather than completed it. The engine was required and configured (`config.active_storage.service` in all three environments) but the original `create_active_storage_tables` migration was never present in this repo — only orphaned follow-up amendment migrations with nothing to amend. Nothing in `app/` uses `has_one_attached`, `has_many_attached`, or direct uploads. `POST /rails/active_storage/direct_uploads` was therefore a public, unauthenticated write endpoint for a feature the app doesn't have, 500ing on every call (Sentry API-8D).

Reinstating it when #53 (photo uploads) is actually built is a single generator invocation (`active_storage:install`), so there's no reason to keep a broken endpoint live to reserve a feature nobody has started on. Linking #53 here so this choice is discoverable when photo upload is picked up.

## Changes

- `config/application.rb`: drop `require 'active_storage/engine'`.
- `config/environments/{production,development,test}.rb`: drop the `config.active_storage.service` lines.
- Delete the four dead `db/migrate/*.active_storage.rb` amendment migrations, and the now-orphaned `config/storage.yml`.
- Drop the dangling `.rubocop_todo.yml` Exclude entry for one of those migrations.
- `db/schema.rb` already had zero `active_storage_*` tables, so no change was needed there — `bundle exec rails db:migrate` from scratch was verified to still succeed and leave the schema consistent.
- New request spec (`spec/requests/active_storage_removed_spec.rb`) pinning that `POST /rails/active_storage/direct_uploads` now returns 404, not 500.

## Verification

- Reproduced the 500 against the pre-fix code, confirmed the new spec catches it.
- Full RSpec suite: 1138 examples, 0 failures.
- RuboCop: 453 files, 0 offenses.

Touches: `config/application.rb`, `config/environments/`, `config/storage.yml`, `db/migrate/`, `.rubocop_todo.yml`, `spec/requests/`


lambda2:
🔎 claim-review-pr 2026-09-14T08:53:00Z — reviewing this PR now.

lambda2:
✅ Reviewed — Pass.

**Contract** (#365 acceptance criteria):
- `POST /rails/active_storage/direct_uploads` no longer 500s, pinned by a request spec — verified: `spec/requests/active_storage_removed_spec.rb` passes, asserts 404.
- `bundle exec rails db:migrate` from scratch succeeds with a schema consistent with actual usage — verified independently: dropped/recreated `trefle_test`, ran `db:migrate` from empty, exit 0, `ActiveRecord::Base.connection.tables.grep(/active_storage/)` → `[]`.
- PR states the decision (removal, not completion) and links #53 for discoverability — present in the PR body.

**CI**: Brakeman, RSpec (1138 examples), RuboCop all green (`gh pr checks 371`).

**Local verification** (fresh worktree):
- `bundle exec rubocop` — 453 files, 0 offenses.
- Targeted spec passes.
- `grep -rn "active_storage\|ActiveStorage\|has_one_attached\|has_many_attached\|direct_upload"` across `app/`, `config/`, `lib/`, `spec/`, `db/schema.rb` — only hits are the new spec itself. No dangling references.
- No explicit `active_storage` route mount outside the removed engine require; the one Gemfile hit is a pre-existing commented-out line, untouched.
- Diff scope matches the PR's `Touches:` line exactly — nothing outside it.

Note: hit an unrelated `LoadError: cannot load such file -- sassc` on `management_spec.rb` in my scratch worktree (missing native gem in that throwaway checkout) — confirmed via `log/test.log` it's an environment artifact, not caused by this diff, and CI's own RSpec run is green at 0 failures.

Moving back to ready for review.

Issue #374, opened 2026-09-14
Default gbif_score sort has no id tiebreaker — pagination can skip/duplicate rows


lambda2:
Found while implementing #367 (deep OFFSET pagination perf work) — filed separately per that issue's own guidance ("if it does, it is worth its own issue"), not fixed there.

## The bug

`Api::ApiController#apply_sort`'s default-sort branch orders by `gbif_score` alone, with no tiebreaker:

```ruby
# app/controllers/api/api_controller.rb
def apply_sort(collection, orderable_fields, default_sort:)
  if params[:order].is_a?(ActionController::Parameters)
    collection.sort_with(permit_known_query_keys(:order, orderable_fields))
  else
    collection.order(default_sort)   # <- no tiebreaker
  end
end
```

This is the path both `Api::V1::PlantsController#collection` and `Api::V1::SpeciesController#collection` use whenever the client doesn't pass an explicit `order` param (`default_sort: { gbif_score: :desc }`) — i.e. it's the default, most-used listing order for `/plants` and `/species`.

Compare `Sortable#sort_with`, used for the *explicit*-`order` path, which already merges an `id: :asc` tiebreaker:

```ruby
# app/models/concerns/sortable.rb
results.order(sort_hash.merge(id: :asc))
```

`gbif_score` is far from unique — a large share of rows share the default value. Postgres does not guarantee a stable row order for ties across separate `OFFSET/LIMIT` queries (plan choice, concurrent writes, autovacuum can all reorder them between requests). A client paging through `/plants` or `/species` with the default sort can silently skip rows that tie with a boundary row on one page and get duplicates of others — with no error, just wrong data.

## Scope

- Add an `id` (or similar) tiebreaker to the default-sort branch in `apply_sort`, matching what `sort_with` already does for explicit sorts.
- Check whether this also needs to interact with the deep-OFFSET work from #367 (whatever landed there — a max page depth guard as of that PR) or with #217 (client page size) if either lands a cursor-based pagination scheme in the meantime — a tiebreaker is a prerequisite for correct keyset pagination either way.
- Add a regression test that orders by a non-unique field across two pages and asserts no row is skipped or duplicated.

## Acceptance criteria

- `/plants` and `/species` with no explicit `order` param produce a total ordering (score, then id) so paging is stable.
- A regression spec demonstrates the fix (e.g. seed several species with an identical `gbif_score`, page through with `limit`/`page`, and assert every id appears exactly once across pages).

[...]

Issue #217, opened 2026-09-03
Client-controlled page size on collection endpoints


lambda2:
Most-requested DX feature on the tracker (community reports: #132, #46). Collection endpoints are fixed at 20 items/page. Meanwhile `/api/v1/species/search` already accepts an **uncapped** `limit` param (`(params[:limit] || 20).to_i`) — a client can request the whole table in one page.

Scope:
- Accept `limit` (integer, 1..100, default 20) on every /api/v1 collection endpoint, wired into pagy `items:`. Out-of-range or non-integer values → 400 in the error envelope (consistent with #205/#216 semantics).
- Cap the existing search `limit` to the same 1..100 range (robustness fix).
- Pagination links (first/prev/next/last) must propagate the param.
- Document in swagger + docs pagination guide.

The PR should close #132 and #46.

Acceptance criteria:
- Request specs: `limit=100` returns 100 items, `limit=0`/`limit=1000`/`limit=abc` return 400, links carry the param.
- Search endpoint capped identically.
- Full suite green.

Touches: app/controllers/api/, app/controllers/concerns/, lib/schemas/, spec/requests, swagger/
Blocked by: #216 (same controller zone)

PR #375, opened 2026-09-14
Cap deep-OFFSET pagination and index lower(scientific_name) (#367)

### LINKED ISSUE #367: Deep OFFSET pagination on gbif_score, and a missing index on lower(scientific_name)
The three slowest queries in production are all `ORDER BY species.gbif_score DESC` with a `LIMIT/OFFSET`, and together they are the highest-volume performance signal the project has — 61,899 grouped events, the oldest running since August 2024.

| Sentry | Transaction | Events | First seen |
|---|---|---|---|
| [API-3Q](https://trefle.sentry.io/issues/API-3Q) | `Api::V1::PlantsController#index` | **59,278** | 2024-08-21 |
| [API-5J](https://trefle.sentry.io/issues/API-5J) | `Api::V1::SpeciesController#index` | 2,286 | 2025-12-31 |
| [API-4M](https://trefle.sentry.io/issues/API-4M) | `Api::V1::SpeciesController#index` | 335 | 2025-06-23 |

```sql
-- API-3Q
SELECT "species".* FROM "species" WHERE "species"."main_species_id" IS NULL
  ORDER BY "species"."gbif_score" DESC LIMIT $1 OFFSET $2
-- API-5J
SELECT "species".* FROM "species" ORDER BY "species"."gbif_score" DESC LIMIT $1 OFFSET $2
-- API-4M
SELECT "species".* FROM "species" WHERE (lower(scientific_name) IN ($1))
  ORDER BY "species"."gbif_score" DESC LIMIT $2 OFFSET $3
```

## These are two different problems

**API-3Q and API-5J: deep OFFSET, not a missing index.** The indexes already exist — `species_gbif_score_idx` on `gbif_score` and `index_species_on_main_species_id_and_gbif_score` on the exact `(main_species_id, gbif_score)` pair API-3Q filters and sorts on. Postgres still has to walk and discard every row before the offset, so page 1 is fast and page 5,000 is not, on a 489k-row table. Volume matters here: `/plants` is the endpoint the public API and the cache warmer both page through exhaustively.

**API-4M: a genuinely missing index.** `lower(scientific_name)` cannot use `species_scientific_name_index`, which is on the raw column. Every lookup by name is a sequential scan. This one is a one-line migration.

## Scope

- **Add the functional index** `ON species (lower(scientific_name))` and confirm with `EXPLAIN` that the API-4M query uses it. Ship it `CONCURRENTLY` — the table is large and this runs against production.
- **Decide what to do about deep OFFSET**, and record the decision in the PR rather than guessing at an implementation:
  - a keyset/cursor page (`WHERE gbif_score < $last` ordered by `(gbif_score, id)`) is the real fix, but it changes the pagination contract and interacts with #217 (client-controlled page size) — it may belong there instead;
  - a maximum page depth, returning a documented `400` past it, is small and defensible for a public API;
  - or bound it at the cache layer, since the warmer is a known heavy pager.
  Pick one, say why, and open a follow-up for the rest.
- `gbif_score` is not unique, so any ordering work must add a tiebreaker (`id`) or pagination silently skips and repeats rows across pages. Check whether that bug exists today before changing anything — if it does, it is worth its own issue.
- Measure before and after on a realistic dataset. "Should be faster" is not an acceptance criterion.

## Acceptance criteria

- `EXPLAIN (ANALYZE)` for the API-4M query shows an index scan on the new functional index, with the before/after timings in the PR body.
- The migration is reversible and safe to run on a 489k-row table without taking a write lock.
- Whatever is chosen for deep OFFSET is implemented **or** explicitly deferred to a linked follow-up issue — not left silent.
- No change to the documented pagination contract without saying so in the PR.

Sentry: `Fixes API-4M`. API-3Q and API-5J close only if deep OFFSET is actually addressed here.

Touches: `db/migrate/`, `db/schema.rb`, `app/controllers/api/v1/species_controller.rb`, `app/controllers/api/v1/plants_controller.rb`, `app/models/concerns/scopes/species.rb`
Blocked by: #364 (same controller zone; take after)

lambda2:
Closes #367

## Two different problems, same three slow queries

Sentry's highest-volume performance signal (61,899 grouped events, oldest since August 2024), all `ORDER BY species.gbif_score DESC LIMIT/OFFSET` on `/plants` and `/species`.

### API-4M — missing index (fixed)

`lower(scientific_name) IN (...)` couldn't use the existing unique index on the raw column, so every name-filtered lookup was a sequential scan. Added a `CONCURRENTLY` functional index on `species (lower(scientific_name))`.

**`EXPLAIN (ANALYZE)`, measured on a full production-size restore (489,358 species rows):**

Before — `Index Scan Backward using species_gbif_score_idx`, `Filter: (lower(scientific_name) = ...)`, `Rows Removed by Filter: 489357`:
```
Execution Time: 464.964 ms
```

After the migration (index build + `ANALYZE`, both run by the migration itself):
```
Limit  (cost=8.45..8.46 rows=1 width=4193) (actual time=0.081..0.082 rows=1 loops=1)
  ->  Sort (Sort Key: gbif_score DESC)
        ->  Index Scan using index_species_on_lower_scientific_name on species
              Index Cond: (lower((scientific_name)::text) = 'eucalyptus robertsonii'::text)
Execution Time: 0.131 ms
```
**465ms → 0.131ms.**

One trap worth flagging explicitly: `CREATE INDEX CONCURRENTLY` doesn't update planner statistics, and autovacuum's analyze threshold is a row-modification count, not triggered by a new index. Right after building the index but *before* `ANALYZE`, the planner still picked the *old* plan against the *new* index (no stats on the expression yet) — same query took **60.8s**. That's why this migration uses explicit `up`/`down` instead of `change`: it runs `execute 'ANALYZE species'` right after `add_index`, which isn't auto-reversible.

Migration is `disable_ddl_transaction!` + `algorithm: :concurrently` on both `up` and `down` — no `ACCESS EXCLUSIVE` lock in either direction, verified with `db:migrate` / `db:rollback` / `db:migrate` locally.

### API-3Q / API-5J — deep OFFSET (capped, not eliminated)

Not a missing index — `species_gbif_score_idx` and `index_species_on_main_species_id_and_gbif_score` already exist. Postgres just has to walk and discard every row before a deep `OFFSET` on a 489k-row table.

**Decision: a maximum page depth, not keyset pagination.** Real cursor/keyset pagination (`WHERE (gbif_score, id) < (?, ?)`) is the actual fix, but it replaces page-number links with an opaque cursor — a change to the public pagination contract that overlaps #217 (client-controlled page size). Doing both as separate contract changes is worse than doing one; #217 is the better home for that. Until it lands, `WithCachedCount::MAX_PAGE_DEPTH` (2,500 pages × the current fixed 20 items/page = a 50,000-row OFFSET ceiling) rejects deeper requests with a documented `400` instead of letting them degrade silently.

**Timings backing the 2,500 cap** (same restore, `/plants` default sort):
| offset | time |
|---|---|
| 0 (page 1) | 0.032ms |
| 49,980 (page 2,500 — the cap) | 26.8ms |
| 200,000 (rejected today) | 109.5ms, still climbing |

**Pagination contract change, documented:** `page` past 2,500 now returns `400` instead of a (very slow) page of results, on `/api/v1/plants` and `/api/v1/species` (including the zone/genus-nested variants, which share the same controller action). Updated the rswag parameter docs for all four affected endpoints to say so. The error message states the limit and points at narrowing with a filter/range instead.

### Tiebreaker bug — found, not fixed here

The issue asked to check whether `gbif_score`'s non-uniqueness causes pagination to skip/repeat rows under the default sort. It does: `Api::ApiController#apply_sort`'s default-sort branch (`collection.order(gbif_score: :desc)`) has no `id` tiebreaker, unlike `Sortable#sort_with`'s explicit-`order` path, which already merges one in. Per the issue's own instruction ("if it does, it is worth its own issue"), **not fixed in this PR** — filed as #374.

### A deliberate deviation from `Touches`

The depth guard lives in `app/controllers/concerns/with_cached_count.rb` — included only by `PlantsController` and `SpeciesController` — rather than in those two controllers or `Scopes::Species` as listed. `scientific_name` already routes through the `lower(...)` branch of `filter_by_scientific_name` with no scope change needed, and hanging the depth check off the shared pagy concern avoids duplicating it across both controllers, which felt like the more faithful reading of the existing pattern (that concern already exists specifically to customize `pagy_get_vars` for these two controllers).

### Other notes from review

- The existing `Pagy::OverflowError`/`VariableError` → 404 path and this new depth guard → 400 path both mean "you paged too far," via different codes. Kept them distinct on purpose: overflow means *this collection doesn't have that many pages*; the depth guard means *we won't run that query regardless of how many pages exist*. Different failure, different code.
- The depth check runs before `cache_count`, so a request past the cap never even issues the (cached, but real) `COUNT` query.

## Testing

- `bundle exec rspec` — 1,157 examples, 0 failures.
- `bundle exec rubocop` — 0 offenses.
- `bundle exec brakeman` — 0 warnings.
- Migration tested both directions (`db:migrate` → `db:rollback` → `db:migrate`) on a fresh restore of a production-size dataset.

Touches: `db/migrate/`, `db/schema.rb`, `app/controllers/api/api_controller.rb`, `app/controllers/concerns/with_cached_count.rb`, `spec/`


lambda2:
🔎 claim-1789385422-5753

lambda2:
## Review: Pass ✅

**Contract checked against:** #367 (acceptance criteria all met) + the deferred/related items it names (#374 tiebreaker bug filed as instructed, #217 overlap correctly punted for the keyset-pagination alternative).

**CI:** `gh pr checks 375` — Brakeman, RSpec, RuboCop all green.

**Local verification** (temp worktree off `worktree-issue-367-gbif-score-pagination`, full production-config services):
- `bundle exec rspec` — 1,157 examples, 0 failures (matches PR body).
- `bundle exec rubocop` — 0 offenses.
- Migration: `db:migrate` → `db:rollback` → `db:migrate`, all clean; rollback took 0.0052s (no `ACCESS EXCLUSIVE` stall), forward migration ran `add_index CONCURRENTLY` (0.96s) + `ANALYZE species` (0.29s) as described.
- Traced `pagy_get_vars` → confirmed `PlantsController#index`/`SpeciesController#index` both route through `#collection`, which handles the `genus_id`/`plant_id`/`zone_id` nested cases before calling `pagy(@collection)` — so the depth guard in `WithCachedCount` (rather than per-controller) does cover all four documented endpoints, matching the PR's stated deviation from `Touches:`.
- Confirmed `search`/`full_search` use `pagy_searchkick` → `pagy_searchkick_get_vars`, a *different* Pagy hook that never touches `WithCachedCount#pagy_get_vars` — so leaving `species/search`'s `page` doc untouched is correct, not an oversight.
- `rescue_from WithCachedCount::PageDepthExceededError` is appended after the existing `Pagy::VariableError` line, consistent with the file's documented most-specific-declared-last convention; `error_code_for(:bad_request)` → `"bad_request"` matches the new request specs.
- Boundary logic (`> MAX_PAGE_DEPTH`, checked before `cache_count` so the COUNT query is skipped) matches the new unit spec's exact-boundary/one-past-boundary pair.

**Acceptance criteria from #367**, all satisfied:
- EXPLAIN (ANALYZE) before/after + index-scan confirmation in the PR body.
- Migration reversible, no write lock — verified above.
- Deep-OFFSET explicitly addressed (capped) with the tradeoff vs. keyset pagination explained and cross-referenced to #217.
- Pagination contract change (400 past page 2,500) documented in the PR body and in the rswag parameter docs for all four affected endpoints.
- Tiebreaker bug investigated, confirmed present, filed as #374 rather than silently left — matches the issue's own instruction.

No gaps against the contract. Label `needs-fixes` not applicable.

PR #376, opened 2026-09-14
Ignore the two crawler spec paths that could reach this public repo

lambda2:
`.gitignore` already covers `lib/crawlers`, `app/workers/crawlers` and `spec/workers/crawlers`. It does not cover `spec/lib/crawlers` or `spec/fixtures/try`, and both are reachable by an ordinary mistake.

## How the mistake happens

The private crawlers repo has no Gemfile. Its specs split in two:

- **worker specs** (`try_import_spec.rb` and friends) `require 'rails_helper'` and need the models and a database, so the private `.rspec` excludes them — they can only be run by copying them **here**. That is why `spec/workers/crawlers` is already ignored.
- **parser specs** (`reader_spec.rb`, `mapper_spec.rb`) and their TRY fixture run standalone over there against this repo's bundle.

Nobody *should* copy the second group here. But `spec/lib/` is a live tracked directory — it holds the checks, ingester, resolver and utils specs — so if anyone does, `git add -A` stages them and private crawler code enters a public repository. Same for `spec/fixtures/try`.

## The change

Two path lines, plus a comment explaining which specs get copied here and why the others must not.

## Verification

Copied a parser spec, the TRY fixture and a worker spec into a clean worktree:

- `git status` reports only `.gitignore`.
- `spec/lib/ingester_spec.rb` stays tracked — the rule is scoped to `spec/lib/crawlers`, not to `spec/lib`.

No application code changed, so the suite is unaffected; CI will confirm.


lambda2:
🔎 claim-1789386920-3046

lambda2:
## Review: Pass ✅

**Contract:** no linked issue — this is a self-justifying hygiene fix per CLAUDE.md's crawler rule (`spec/lib/crawlers`, `app/workers/crawlers`, `datasets/`, `records/` etc. must never be committed). The PR's own rationale is the acceptance criteria: two live-tracked directories (`spec/lib/`, `spec/fixtures/`) can silently swallow copied-in private crawler specs on `git add -A`.

**CI:** `gh pr checks 376` — Brakeman, RSpec (2m41s), RuboCop all green.

**Local verification** (temp worktree off `chore-ignore-crawler-spec-paths`):
- Copied in a fake parser spec (`spec/lib/crawlers/reader_spec.rb`), a TRY fixture (`spec/fixtures/try/...`), and a worker spec (`spec/workers/crawlers/try_import_spec.rb`) — none showed up in `git status`, confirming all three ignore lines work.
- `spec/lib/ingester_spec.rb` (an unrelated, already-tracked file under the same `spec/lib/` parent) stayed tracked — the new `spec/lib/crawlers` rule is correctly scoped and doesn't shadow sibling specs, matching the PR's stated verification.
- Confirmed master's current .gitignore only has `spec/workers/crawlers` (line 101), matching the PR's 'before' description.
- Grepped the tracked tree for any reference to `spec/lib/crawlers`, `spec/fixtures/try`, or `spec/workers/crawlers` in Ruby/YAML/Markdown — nothing depends on these paths being tracked (no coverage config, no CI step, no doc), so ignoring them is safe.

No application code changed, diff is two .gitignore lines plus a comment. No gaps against the stated rationale.

Issue #26, opened 2020-01-23
Ability to add data


GreenThumbAU:
I'm using the API to prefill data for plants in our app's admin, when we add/create them. There are a few that Trefle returns as empty [] or complete_data = false and I'm wondering if you're considering adding the ability to add or update data via the API. That way, when we complete adding a plant to our database, we could also add it to Trefle or update incomplete data. Thoughts?

lambda2:
That's planned for the next release !

PR #388, opened 2026-09-16
Resolve TRY's per-dataset credit: 727 numbers in a free-text field become names

### LINKED ISSUE #387: TRY's per-dataset credit is 727 numbers in a free-text field: build the register that resolves them
TRY is published under CC BY, and its release notes ask for two citations: the TRY reference itself, and the reference of **each contributing dataset**. The first is carried on the `ForeignSource` and reaches the API through `sources[].citation`. The second is not carried at all.

A fact names its datasets, but only as a free-text list in `notes`:

```
TRY 7.0 datasets 1, 4, 7, 25, 71, 86, 92, 110, 175, 178, 193, 203, 236, 443, 460, 818, 879, 882, 918, 925
```

Twenty datasets, for one attribute on one species. The numbers are exposed — `/api/v1/species/:id/facts` serializes `notes` — but nothing resolves them to a person. As credit, it is inert.

The decision to carry this through a resolvable register rather than inlining every citation is recorded on [treflehq/trefle-crawlers#26](https://github.com/treflehq/trefle-crawlers/issues/26#issuecomment-5684536561). This issue is the public-side half: the register itself.

## Measured

Extracted from the four TRY 7.0 exports (columns `DatasetID`, `Dataset`, `LastName`, `FirstName`, `Reference`):

| | |
|---|---:|
| Distinct datasets | **727** |
| Distinct contributors | **508** |
| (dataset, reference) pairs | 810 |
| Datasets with **no** citable reference | **139** (19%) |

Three things worth knowing before building it:

- **It is one-to-many.** 810 pairs for 727 datasets — dataset 1 alone cites three publications. Keying the register on `dataset_id` collapses them.
- **139 datasets ship no publication reference** (`unpub.` or empty). They are still owed credit, and the export carries a contributor name and a dataset name for each, so those carry it. Per this repo's existing rule: do not invent a citation.
- **The exports contain non-UTF-8 bytes** in the reference strings — `cut` fails outright on them without `LC_ALL=C`. The register must be transcoded when generated, not read hopefully as UTF-8 later.

## Scope

- A `try_datasets` table: `dataset_id`, `dataset_name`, `contributor`, `reference` (nullable), unique on `(dataset_id, reference)`.
- The register committed as data under `db/data/`, not derived at boot. **The exports are 30 GB and exist on one laptop** — the credit has to outlive them. Note `*.csv` is gitignored repo-wide to keep bulk crawler exports out of a public checkout, so the register needs a committed format that does not fight that rule rather than an exception to it.
- A replayable seeder on the `app/workers/migrators/` pattern. It must never delete: a dataset dropping out of a later export does not un-owe its credit for values already ingested from it.

## Acceptance

- Every dataset id a fact cites resolves to a row. Measured on the dev import: 254 of the 255 ids cited across a 20,000-fact sample already resolve against the extracted register — the one miss is `0`, an artefact of scanning `\d+` over the string `TRY 7.0 datasets …`, which is itself an argument for a structured column instead of free text.
- A dataset with no reference still yields a credit line from its contributor and dataset name.
- Re-seeding is a no-op.

## Follow-up, not this issue

- Exposing the register: linking a fact to its datasets in the API, and publishing the register page alongside the sources register.
- Replacing the free-text `notes` list with a structured column — the `0` above shows why.

Blocks the honest completion of #328/#382: promotion is what makes TRY-derived values visible in the main payload.

lambda2:
Closes #387. The decision behind it is on [treflehq/trefle-crawlers#26](https://github.com/treflehq/trefle-crawlers/issues/26#issuecomment-5684536561).

TRY is CC BY, and its release notes ask for two citations: the TRY reference, and the reference of **each contributing dataset**. The first reaches the API through `sources[].citation`. The second existed nowhere a reader could use — a fact names its datasets only as free text:

```
TRY 7.0 datasets 1, 4, 7, 25, 71, 86, 92, 110, 175, 178, 193, 203, 236, …
```

Twenty datasets, one attribute, one species. The numbers are exposed; nothing resolved them to a person.

## What is here

A `try_datasets` table, the register committed under `db/data/`, a replayable seeder, and `TryDataset#credit`, which returns the reference when there is one and the contributor + dataset name when there is not.

**727 datasets, 508 contributors, 810 pairs, 139 without a citable reference.**

## Three things the data decided, not the design

**It is one-to-many.** 810 pairs for 727 datasets — dataset 1 cites three publications. Keyed on `(dataset_id, reference)`; keying on `dataset_id` alone silently drops two thirds of dataset 1's credit. Pinned by a spec, verified by breaking the key and watching that spec fail.

**139 datasets ship no reference** (`unpub.` or empty). Still owed credit, so `#credit` falls back to what the export does carry. Not dropped, not invented — this repo's existing rule.

**The exports carry non-UTF-8 bytes** in the reference strings; `cut` fails outright on them without `LC_ALL=C`. Transcoding happens when the register is generated. The reader takes UTF-8 and lets a bad byte fail loudly rather than seeding mojibake into an attribution.

## Two choices worth flagging

**Committed as data, not derived at boot.** The exports are 30 GB and exist on one laptop. The credit has to outlive them.

**YAML, not CSV.** `*.csv` is gitignored repo-wide to keep bulk crawler exports out of a public checkout. This file is the opposite kind of thing — the credit CC BY obliges us to publish — so it takes a committed format rather than an exception to that rule. I did not force it past `.gitignore`.

## Verification

Seeded against the real import: of the **255 dataset ids cited across a 20,000-fact sample, 254 resolve**. The one miss is `0` — an artefact of scanning `\d+` over the string `TRY 7.0 datasets …`, not a gap in the register. That is itself the argument for replacing the free-text list with a structured column, which is follow-up on #387 rather than scope here.

RSpec **1290 examples, 0 failures** (coverage 81.60%) · RuboCop clean on tracked files · migration is additive (new table, two indexes).

## Not in this PR

Exposing the register through the API and publishing it alongside the sources page. The table and its data have to exist first; both are listed as follow-up on #387.


lambda2:
🔎 claim-1789546455-32734

lambda2:
🔎 Review (post-merge — this landed seconds after the claim comment above, apparently picked up by merge-train before the review gate applied):

Checked against #387's acceptance criteria:
- One-to-many keying on `(dataset_id, reference)` — correct per the issue's dataset-1-cites-three-publications case.
- 139 no-reference datasets fall back to `contributor — dataset_name` via `TryDataset#credit` rather than dropping or inventing a citation.
- Seeding is idempotent and non-destructive (`TryDatasetRegister.apply`): replay is a no-op, a dataset absent from a newer export keeps its credit — both directly spec'd.
- UTF-8 transcoding is pushed to register-generation time; the reader trusts UTF-8 and fails loudly on bad bytes.
- Committed YAML (not CSV) is consistent with this repo's crawler-artifact `.gitignore` rule, with the rationale documented inline.
- Spec validates the committed register's shape (810 rows / 727 datasets / 139 blank references), matching the PR description's numbers.

CI (Brakeman, RSpec, RuboCop) was green. No correctness gaps found against the issue's acceptance criteria.

PR #389, opened 2026-09-16
Release 2.7.0

lambda2:
Six commits since `v2.6.0`. Three change how the database behaves rather than what the API answers.

| | |
|---|---|
| **#382** | `Migrators::FactPromotion` — fills an empty trait column from the strongest sourced fact; `Checks::FactDivergence` reports a filled column a fact contradicts. **Dry-run by default**, nothing writes unless asked. |
| **#388** | The TRY dataset register — 727 datasets, 508 contributors. Carries **the one migration** in this release (`create_table :try_datasets`), additive. |
| **#386** | TRY's licence recorded as CC-BY-4.0. Null meant *not verified*, the wrong claim for a source that states its terms. |
| **#383** | The sweep stops walking past the pagination cap 2.6.0 introduced. **This is live noise today**: ~66 expected 400s every six minutes in production, each reported to Sentry as a sweep failure. |
| **#384** | The source priority table listed World Flora Online twice. |
| **#380** | The runbook's image tags, which carried a `v` the registry does not. |

The spec is regenerated for the version bump — the drift guard added in 2.6.0 fails CI otherwise. First time it has had something to catch.

## Migration

One, additive: `create_table :try_datasets` plus two indexes. Nothing is dropped or renamed, so the canary flow in `DEPLOYMENT.md` applies unchanged — the canary will run it against the production database on boot, as designed.

## After the tag

The image publishes as `ghcr.io/treflehq/api:2.7.0` — **no `v`**, per #380.

Worth checking a 20-minute window of `trefle-api` logs after the switch: #383's acceptance criterion is no sweep-originated 400s, and the sweep runs every 20 minutes, so one cycle tells.

RSpec 1290 examples, 0 failures · RuboCop clean on tracked files.


lambda2:
🔎 claim-1789547206-19257

lambda2:
🔎 Review:

Diff against `master` is exactly the version bump: `config/application.rb` VERSION `2.6.0` → `2.7.0`, and `public/swagger/v1/swagger.yaml` `info.version` matching. Nothing else changed — the six functional commits described in the PR body (#382, #388, #386, #383, #384, #380) are already on `master`; this branch adds only the release commit on top.

- Both version strings bumped in lockstep — the CI drift guard (`.github/workflows/ci.yml:86-89`, regenerate via `rake rswag:specs:swaggerize` and diff) would fail otherwise, and it's green here.
- Migration described (`create_table :try_datasets`) is additive-only and already landed on `master` via #388, consistent with the canary-safe claim.
- No crawler/dataset/deploy paths touched, nothing gitignored got dragged in.

CI (Brakeman, RSpec, RuboCop) green. No correctness gaps against the stated contract (clean version bump for the tag).

